PR #9, opened 2025-07-02
fix(README): Rename LED -> Led for Quick Start section

Jon-Liang:
Following the README as-is results in the following error

`Error: Failed to resolve module path '@github/diodeinc/stdlib:main/generics/LED.star': File not found`

and requires renaming reference to LED.star -> Led.star and variable names renamed accordingly for consistency.


LK (approved):
Thanks Jon :)

PR #19, opened 2025-07-06
[MODULE] resolve workspace loads

hexdae:


crates/pcb-zen/tests/module_loading.rs
```diff
@@ -1,68 +1,258 @@
...
+
+    // Evaluate the test file
+    let result = ctx
+        .set_source_path(test_file.clone())
+        .set_module_name("test".to_string())
+        .set_inputs(InputMap::new())
+        .eval();
+
+    // Check if there were any errors
+    for diag in &result.diagnostics {
```
LK: These tests might be nicer as snapshot tests?

docs/spec.md
```diff
@@ -57,7 +58,7 @@ load("@github/user/repo:branch/path.zen", "function")
 load("@gitlab/user/repo:branch/path.zen", "function")
 
 # GitLab repository (nested groups)
-load("@gitlab/kicad/libraries/kicad-symbols:v7.0.0/Device.kicad_sym", "Resistor")
+Symbol(library = "@gitlab/kicad/libraries/kicad-symbols:v7.0.0/Device.kicad_sym", name = "R_US")
```
LK, line 61: Nice, ty

docs/spec.md
```diff
@@ -92,8 +93,6 @@ my-lib = "@github/myorg/mylib:v1.0.0"
 local-lib = "./path/to/local/lib"
 ```
 
-## Core Types
```
LK: intentional?

crates/pcb-zen-core/src/lib.rs
```diff
@@ -465,6 +465,13 @@ impl LoadResolver for WorkspaceLoadResolver {
         load_path: &str,
         current_file: &Path,
     ) -> Result<PathBuf, anyhow::Error> {
+        // Reject paths that look like package imports - these should be handled by RemoteLoadResolver
+        if load_path.starts_with('@') {
```
LK: Do we need this?

crates/pcb-zen-core/src/lang/module.rs
```diff
@@ -640,7 +653,18 @@ pub fn module_globals(builder: &mut GlobalsBuilder) {
         // Resolve the path using the load resolver
         let resolved_path = load_resolver
             .resolve_path(file_provider.as_ref(), &path, current_file)
-            .map_err(|e| anyhow::anyhow!("Failed to resolve module path '{}': {}", path, e))?;
+            .map_err(|e| {
+                if path.starts_with('@') {
+                    anyhow::anyhow!(
+                        "Failed to resolve module path '{}': {}. \
+                        This might indicate that the load resolver doesn't support remote package imports. \
+                        Ensure you're using the proper evaluation context setup.",
```
LK: I'm not a huge fan of this error since it's talking about internal implementation details of the interpreter (so it's not really actionable for the user)

crates/pcb-zen/tests/module_loading.rs
hexdae: yes you are right, I followed the path in the file, but I would like to convert all of these to snapshots in a follow up

docs/spec.md
hexdae: no, will fix

crates/pcb-zen-core/src/lang/module.rs
hexdae: yeah I don't like it either

crates/pcb-zen-core/src/lib.rs
hexdae: nah, let's get rid of this

crates/pcb-zen/tests/module_loading.rs
LK: You can use the original test as an example: https://github.com/diodeinc/pcb/blob/main/crates/pcb-zen/tests/module_loading.rs

or pcb-zen-core has a more succinct macro for this that I've been meaning to port over to this crate

hexdae: updated

PR #25, opened 2025-07-10
[FMT] implement formatter

hexdae:
Add a built in formatter to `pcb` based on starlark's buildifier

### DIFF
--- a/crates/pcb/Cargo.toml
+++ b/crates/pcb/Cargo.toml
@@ -31,4 +31,4 @@ pcb-kicad = { workspace = true }
 open = { workspace = true }
 inquire = { workspace = true }
 pcb-ui = { workspace = true }
-pcb-buildifier = { path = "../pcb-buildifier" }
+pcb-buildifier = { workspace = true }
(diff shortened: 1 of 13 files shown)


crates/pcb/Cargo.toml
```diff
@@ -30,4 +30,5 @@ colored = { workspace = true }
 pcb-kicad = { workspace = true }
 open = { workspace = true }
 inquire = { workspace = true }
-pcb-ui = { workspace = true } 
+pcb-ui = { workspace = true }
+pcb-buildifier = { path = "../pcb-buildifier" }
```
LK: Can we do the same workspace thing we do for the other deps?

crates/pcb/src/fmt.rs
```diff
@@ -0,0 +1,187 @@
...
+        Ok(true)
+    }
+}
+
+/// Collect .zen and .star files from the provided paths
+pub fn collect_files(paths: &[PathBuf]) -> Result<Vec<PathBuf>> {
+    let mut unique: HashSet<PathBuf> = HashSet::new();
+
+    if !paths.is_empty() {
+        // Collect files from the provided paths (non-recursive)
```
LK: should we do this recursively like "cargo fmt" does? or maybe better to match the behavior of build/, but it would be cool to just run this on the root of a repo

LK (approved):

crates/pcb/src/fmt.rs
hexdae: I guess this should work like cargo fmt, so makes sense

PR #50, opened 2025-07-27
Quick & dirty fix for KiCad Python Paths on Windows

jglanz:
# Quick & dirty fix for KiCad Python Paths on Windows
> NOTE: This is quick & DIRTY, better error handling and error messages should be added in addition.  Also, `Vec<Path>` find first should be added checking all provided candidates, etc

- Made `venv` site-packages path resolution more robust by checking `interpreter` parent path
- Added `user` install location for `KiCad` interpreter check


crates/pcb-kicad/src/lib.rs
```diff
@@ -38,19 +48,28 @@ mod paths {
...
 mod paths {
+    use std::path::Path;
+    use crate::path_helpers;
+
     pub(crate) fn python_interpreter() -> String {
-        std::env::var("KICAD_PYTHON_INTERPRETER")
-            .unwrap_or_else(|_| r"C:\Program Files\KiCad\9.0\bin\python.exe".to_string())
+        let paths = vec![
+            std::env::var("KICAD_PYTHON_INTERPRETER")
+                .unwrap_or_else(|_| "python.exe".to_string()),
```
LK, line 67: Would this default to using the system `python.exe` if `KICAD_PYTHON_INTERPRETER` isn't defined?

LK (commented):
Thanks for the PR!

crates/pcb-kicad/src/lib.rs
jglanz: The actual interpreter will/should run from the PATH, but the `site-packages` functions will not recognize the directory structure.
 
Personally, I'd create a fallback `fn` to scan the path (equivalent of`which`) would facilitate what you are suggesting in a more complete way.

jglanz: I just updated the PR to resolve a compilation issue the windows fix caused the other build targets

PR #51, opened 2025-07-28
add nix flake

0xcaff:
```
nix profile install github:diodeinc/pcb
```

or to test

```
nix run github:0xcaff/pcb/martin/nix -- --help
```

kicad isn't packaged here yet but can deal with that later

used crane as there is a bug in nixpkgs rust https://discourse.nixos.org/t/rust-with-git-deps-symbolic-link-error-to-cargo-vendor-dep/28841 and naserk had some issue as well. crane isn't ideal because it doesn't cache properly but now it one click installs so whatever. i will go make the rust builds less stupid in nix with dynamic derivations soon and come back and fix this then

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,9 +21,20 @@
         inherit (pkgs) lib;
 
         craneLib = crane.mkLib pkgs;
-        src = craneLib.cleanCargoSource ./.;
+        src = pkgs.lib.cleanSourceWith {
+          src = lib.cleanSource ./.;
+
+          filter = (
+            orig_path: type:
+            pkgs.lib.hasSuffix ".py" (baseNameOf (toString orig_path))
+            || (craneLib.filterCargoSources orig_path type)
+          );
+
+          name = "pcb-source";
+        };
 
         commonArgs = {
+          pname = "pcb";
           inherit src;
           strictDeps = true;
           doCheck = false;
@@ -47,13 +58,7 @@
           commonArgs
           // {
             inherit cargoArtifacts;
-            pname = "pcb";
             inherit (craneLib.crateNameFromCargoToml { inherit src; }) version;
-
-            prePatch = ''
-              substituteInPlace crates/pcb-layout/src/lib.rs \
-                --replace-fail "scripts/update_layout_file.py" ${crates/pcb-layout/src/scripts/update_layout_file.py}
-            '';
 
             doCheck = false;
           }
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,62 @@
+{
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+    crane.url = "github:ipetkov/crane";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs =
+    {
+      self,
+      nixpkgs,
+      crane,
+      flake-utils,
+      ...
+    }:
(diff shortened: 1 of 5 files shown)


0xcaff:
hmm tried it on x86_64-linux and it doesn't work there, let me go fix the upstream rust nix derivations first its getting quite annoying

LK:
This is great, thanks! We also have pre-built binaries for some platforms, would it be easier to vendor those or does Nix prefer to build things from source?

0xcaff:
> This is great, thanks! We also have pre-built binaries for some platforms, would it be easier to vendor those or does Nix prefer to build things from source?

nix prefers build from source for reproducibility. the nix derivation holds the full set of build instructions to build a package including all sub-packages (which in turn include build instructions for themselves and their deps). nix has binary caches of its own which add some hashing + information about what build inputs correspond to what build outputs (https://github.com/0xcaff/ps-nix/blob/1cfda34b130384698769eab86c284d6882c44c9c/flake.nix#L2-L9) but i wouldn't worry about this for now

I should have some time tomorrow to finish this

0xcaff:
ok, this builds on linux now too. didn't bother with getting kicad added to this flake as its broken in upstream nixpkgs on macos and the fix looks non-trivial https://github.com/NixOS/nixpkgs/blob/94def634a20494ee057c76998843c015909d6311/pkgs/applications/science/electronics/kicad/default.nix#L331

flake.nix
```diff
@@ -0,0 +1,74 @@
...
+          commonArgs
+          // {
+            inherit cargoArtifacts;
+            pname = "pcb";
+            inherit (craneLib.crateNameFromCargoToml { inherit src; }) version;
+
+            prePatch = ''
+              substituteInPlace crates/pcb-layout/src/lib.rs \
+                --replace-fail "scripts/update_layout_file.py" ${crates/pcb-layout/src/scripts/update_layout_file.py}
+            '';
```
LK: I'm a bit worried about silently breaking this when we add more runtime deps, is it possible to teach `cleanCargoSource` to keep *.py files instead?

flake.nix
```diff
@@ -0,0 +1,74 @@
...
+          BUILDIFIER_BIN = "${pkgs.bazel-buildtools}/bin/buildifier";
+        };
+
+        cargoArtifacts = craneLib.buildDepsOnly commonArgs;
+
+        pcb = craneLib.buildPackage (
+          commonArgs
+          // {
+            inherit cargoArtifacts;
+            pname = "pcb";
```
LK: When I build this I get a warning about Crane not detecting the name from Cargo.toml (which makes sense since it's just a workspace definition), I think it's coming from the `buildDepsOnly`.

Moving `pname = "pcb"` up to `commonArgs` seems to fix it I think

LK (commented):
Thanks Martin this works great! 2 minor requests

flake.nix
0xcaff: sure we can do that

flake.nix
0xcaff: will update

0xcaff: i guess fixing this is pretty important as that warning shows up on every install

flake.nix
0xcaff: its done

0xcaff:
@LK its ready when you are!

LK:
looks great, thanks @0xcaff!

PR #66, opened 2025-08-14
[DIO-298] [DIO-299] Add `pcb bom` command to generate unmatched bom

akhilles:
To test:

```
pcb bom <path/to/zen> -o json
```

For now, only resistors and capacitors are treated as "well-known" and have more structured data:

```json
  {
    "designators": [
      "C6"
    ],
    "package": "0805",
    "value": "1uF 10% 50V",
    "capacitance": {
      "value": 0.000001,
      "tolerance": 0.1
    },
    "voltage": {
      "value": 50.0
    },
    "dnp": false
  },
  {
    "designators": [
      "R1",
      "R3"
    ],
    "package": "0402",
    "value": "33 1%",
    "resistance": {
      "value": 33.0,
      "tolerance": 0.01
    },
    "dnp": false
  },
```

will add more passives in future PR.


crates/pcb-sch/src/bom.rs
```diff
@@ -0,0 +1,603 @@
...
+            voltage,
+        });
+    }
+
+    // Group entries and return
+    group_bom_entries(bom_entries)
+}
+
+/// Detect well-known modules based on instance.type_ref
+fn detect_well_known_module(
```
LK, line 172: we used to have [`type = "resistor"`](https://github.com/diodeinc/stdlib/blob/bba910b91f6f1320df87147a42b465f876f64690/generics/Resistor.zen#L65) in the generics library but looks like it got lost at some point, this seems like the best we can do for now but we should go back to a state where you explicitly opt-in to generics resolution by specifying what type of component it is IMO (after we update the standard library). not something we need to do today though

crates/pcb-sch/src/bom.rs
```diff
@@ -0,0 +1,603 @@
...
+    #[serde(flatten)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub well_known_module: Option<WellKnownModule>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub voltage: Option<Value<Voltage>>,
+    pub dnp: bool,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+struct GroupKey {
```
LK, line 31: maybe need alternates here too?

crates/pcb/src/bom.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+pub fn write_bom_json<W: Write>(entries: &[BomEntry], writer: W) -> Result<()> {
+    serde_json::to_writer_pretty(writer, entries).context("Failed to write JSON BOM")?;
+    Ok(())
+}
+
+fn write_bom_table<W: Write>(entries: &[BomEntry], mut writer: W) -> Result<()> {
+    let mut table = Table::new();
+    table.load_preset(UTF8_FULL_CONDENSED);
```
LK, line 86: do you know if there's a way to specify a max width on the table?

LK (approved):
great stuff, thanks akhil! couple notes, nothing blocking

crates/pcb-sch/src/bom.rs
akhilles: yes, will add that to stdlib. I need a couple other changes to stdlib as well to finish out this bom feature. e.g. lists and records are being flattened out in in `config.zen`'s `format_value(value)`. so, I'm having to do extra work to parse physical values back from strings.

crates/pcb/src/bom.rs
akhilles: yes, you can set a hard char limit, but I think this might be better:
```
table.set_content_arrangement(comfy_table::ContentArrangement::DynamicFullWidth);
```

it sets the max width to the width of the terminal, so its always formatted nicely. I can also add a hard limit if you prefer.

crates/pcb-sch/src/bom.rs
akhilles: done

crates/pcb-sch/src/bom.rs
akhilles: updated the PR to use the new "type" attribute. also, using the physical values as is, rather than parsing from string.

PR #76, opened 2025-08-15
Read MPN from component properties

LK:
In `stdlib/config.zen` we put the component MPN in to the `properties` field, to be compatible with BOM, etc we should parse this and attach it as the true component MPN.

### DIFF
--- a/crates/pcb-zen-core/src/lang/component.rs
+++ b/crates/pcb-zen-core/src/lang/component.rs
@@ -389,9 +389,20 @@ where
                 );
             }
 
+            // Determine final MPN: explicit kwarg takes precedence; otherwise, use properties["mpn"/"MPN"/"Mpn"] if a string
+            let final_mpn: Option<String> = if let Some(v) = mpn {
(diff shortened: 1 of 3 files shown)


akhilles:
> In `stdlib/config.zen` we put the component MPN in to the `properties` field, to be compatible with BOM, etc we should parse this and attach it as the true component MPN.

Just curious: by "true component MPN", do you mean the `ComponentValue`'s `mpn` field? Is this currently used anywhere? For kicad and the schematic instance attributes, I believe we're using properties.

LK:
> > In `stdlib/config.zen` we put the component MPN in to the `properties` field, to be compatible with BOM, etc we should parse this and attach it as the true component MPN.
> 
> Just curious: by "true component MPN", do you mean the `ComponentValue`'s `mpn` field? Is this currently used anywhere? For kicad and the schematic instance attributes, I believe we're using properties.

sorry you are totally right! I was misinterpreting some other weird behavior I was seeing (which was a bug on the zener side), this shouldn't be necessary.

PR #79, opened 2025-08-17
[FEATURE] sync zones and graphics

hexdae:
- **sync zones and text**
- **add initial test for zone sync**
- **add zones to groups**
- **simpler zone test**
- **fix zone sync**
- **fix patch**


crates/pcb-layout/src/scripts/update_layout_file.py
```diff
@@ -583,6 +631,288 @@ def render_tree(self, indent: int = 0) -> str:
...
+        if not self._outline_points or not self._outline_points[0]:
+            return None
+
+        all_points = [pt for contour in self._outline_points for pt in contour]
+        min_x = min(x for x, y in all_points)
+        max_x = max(x for x, y in all_points)
+        min_y = min(y for x, y in all_points)
+        max_y = max(y for x, y in all_points)
+
+        return VirtualBoundingBox(min_x, min_y, max_x - min_x, max_y - min_y)
```
akhilles: Just curious if `get_kicad_bbox()` would work here? I think it just calls kicad item's `GetBoundingBox()`

crates/pcb-layout/src/scripts/update_layout_file.py
```diff
@@ -583,6 +631,288 @@ def render_tree(self, indent: int = 0) -> str:
...
+
+    def move_by(self, dx: int, dy: int) -> None:
+        """Transform all outline points by offset."""
+        outline = self.kicad_item.Outline()
+        for i in range(outline.OutlineCount()):
+            contour = outline.COutline(i)
+            for j in range(contour.PointCount()):
+                pt = contour.CPoint(j)
+                contour.SetPoint(j, pcbnew.VECTOR2I(pt.x + dx, pt.y + dy))
+        self._capture_outline()  # Update cached points
```
akhilles: I see kicad `Zone` has `move()`:

```py
    def move(self, delta: Vector2):
        """Moves the zone by the given delta vector"""
        self.outline.move(delta)
        for polygon in self.filled_polygons.values():
            for shape in polygon:
                shape.move(delta)
```

not sure if it'll work here

hexdae: yep, added

hexdae:
>I 'm probably missing something but I think there are 2 strategies being used for syncing elements:
clone_to_board + move_by
capture_geometry + apply_geometry
Is there a reason to require both? Would clone_to_board + move_by not suffice?

Initially I thought I needed the capture geometry for some of the polygons, but almost everything we want to move has a .move option, so good idea, removed those methods

PR #80, opened 2025-08-18
Fix package resolution in pcb vendor/release

akhilles:
The TrackingLoadResolver middleware couldn't properly resolve packages.
To fix, eliminate the middleware entirely and enhance CoreLoadResolver
to handle all the necessary tracking. This is likely to reduce future
bugs as well.

@hexdae this should fix the issue you ran into

### DIFF
--- a/test-workspaces/with-pcb-toml/pcb.toml
+++ b/test-workspaces/with-pcb-toml/pcb.toml
@@ -2,4 +2,4 @@
 name = "zener"
 
 [packages]
-registry = "@github/diodeinc/registry"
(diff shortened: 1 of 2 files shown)


test-workspaces/with-pcb-toml/pcb.toml
```diff
@@ -1,2 +1,5 @@
 [workspace]
 name = "zener"
+
+[packages]
+registry = "@github/diodeinc/registry"
```
LK: can we use a public repo for this? (`diodeinc/kicad` maybe?)

PR #89, opened 2025-08-19
Initial implementation of interface subclasses + auto-promotion

akhilles:
ref DIO-297

Should be fully backwards compatible. Adds support for:
- `serialize()` and `deserialize()`
- `using()` to indicate valid promotion targets
- conflict detection in case of multiple promotion targets of the same type
- automatic type promotion across module boundaries
- tracking of promotion targets via a `promotion_by_type` map

demo.zen file that demos the new features: https://gist.github.com/akhilles/2dcfdb0dc2c1049116605d6bfa1a8df5
To test: `pcb build demo.zen`


crates/pcb-zen-core/src/lang/interface.rs
```diff
@@ -448,94 +807,124 @@ impl<'v, V: ValueLike<'v>> DeepCopyToHeap for InterfaceValueGen<V> {
...
-            interface_type_data: OnceCell::new(),
-            fields,
-            param_spec: self.param_spec.clone(),
-        }))
-    }
-}
-
 #[starlark_module]
 pub(crate) fn interface_globals(builder: &mut GlobalsBuilder) {
+    fn using<'v>(value: Value<'v>, eval: &mut Evaluator<'v, '_, '_>) -> anyhow::Result<Value<'v>> {
```
LK, line 812: Should we make this work with `using(Net)` too?

LK (approved):
This is awesome! Can we also update `spec.mdx` with an explanation of the new behavior?

crates/pcb-zen-core/src/lang/interface.rs
akhilles: yeah, probably. updated to allow this.

akhilles:
> This is awesome! Can we also update `spec.mdx` with an explanation of the new behavior?

done!

PR #98, opened 2025-08-21
Use smart rust cache

akhilles:
Out cache sizes are blowing up:

Cache Size: ~8914 MB (9346691088 B)
"C:\Program Files\Git\usr\bin\tar.exe" -xf C:/a/_temp/277318a1-9141-472d-bf1b-4d0461835b0b/cache.tzst -P -C C:/a/pcb/pcb --force-local --use-compress-program "zstd -d" Cache restored successfully
Cache restored from key: Windows-cargo-02b001115c33129b2426bd7db7faa11329497685bfa1d1e328c20e41bc781dca

Which is slowing down builds. Just fetching cargo cache took 6 mins.

Swatinem/rust-cache has much better default behavior:
- Cleans dependencies no longer used
- Cleans incremental build artifacts
- Cleans anything older than 1 week

It also only caches dependencies (and their built artifacts). Caching workspace crates is not that useful.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -115,23 +115,19 @@ jobs:
           cargo run -- vendor test-workspaces/with-pcb-toml
           cargo run -- build test-workspaces/with-pcb-toml/boards --offline
 
+      - name: Checkout stdlib
+        uses: actions/checkout@v5
+        with:
+          repository: diodeinc/stdlib
+          path: stdlib
+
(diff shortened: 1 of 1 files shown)


LK (approved):
wonderful thank you!

LK:
> thanks my computer always runs out of memory

i think this is only for CI but we'll get you a new computer don't worry

PR #101, opened 2025-08-22
[DIO-333] Add `--source-only` releases, `schema_version`

akhilles:


crates/pcb/src/release.rs
```diff
@@ -23,6 +23,14 @@ use crate::workspace::{
     WorkspaceInfo,
 };
 
+const RELEASE_SCHEMA_VERSION: &str = "0.1";
```
LK: thoughts on just a flat version number?
```suggestion
const RELEASE_SCHEMA_VERSION: &str = "1";
```

crates/pcb/src/release.rs
```diff
@@ -64,64 +76,87 @@ pub struct ReleaseInfo {
...
     }
 
-    // Calculate zip path
-    let zip_path = format!("{}.zip", release_info.staging_dir.display());
+    // Execute finalization tasks
+    execute_tasks(&release_info, FINALIZATION_TASKS, using_human)?;
+
+    // Calculate zip path with source suffix if needed
+    let zip_path = if matches!(release_info.kind, ReleaseKind::SourceOnly) {
+        format!("{}.source.zip", release_info.staging_dir.display())
```
LK, line 156: 👌

crates/pcb/src/release.rs
akhilles: done

PR #102, opened 2025-08-22
[DIO-334] Switch to `<board_name>/<version>` tags 

akhilles:


Cargo.toml
```diff
@@ -93,8 +93,6 @@ wasm-bindgen-test = "0.3"
 web-sys = { version = "0.3", features = ["console"] }
 zip = { version = "4.3", default-features = false, features = ["deflate"] }
 flate2 = "1.0.34"
-openssl-sys = { version = "0.9", features = ["vendored"] }
```
LK, line 96: IIRC this was needed at some point because there was a transitive dependency on `openssl`, and if we didn't vendor it then some people had issues installing on linux if they didn't have `openssl-dev` installed

crates/pcb/tests/snapshots/simple__simple_workspace_release.snap
```diff
@@ -7,8 +7,8 @@ Exit Code: 0
 
 --- STDOUT ---
 {
-  "archive": "<TEMP_DIR>/.pcb/releases/boards/TestBoard/unknown.zip",
-  "staging_directory": "<TEMP_DIR>/.pcb/releases/boards/TestBoard/unknown",
+  "archive": "<TEMP_DIR>/boards/.pcb/releases/TestBoard/unknown.zip",
+  "staging_directory": "<TEMP_DIR>/boards/.pcb/releases/TestBoard/unknown",
```
LK, line 11: does this change where we put the `.pcb` folder? I kind of liked there being just one per workspace

Cargo.toml
LK: actually looking at `cargo tree` I don't see `openssl` anymore, so I guess we dropped this dependency, all good!

crates/pcb/src/release.rs
```diff
@@ -345,11 +346,12 @@ fn git_version_and_hash(path: &Path) -> Result<(String, String)> {
         let tags = String::from_utf8(tag_out.stdout)?;
         let tags: Vec<&str> = tags.lines().collect();
 
-        // Use the first tag if any exist
-        if let Some(tag) = tags.first() {
-            if !tag.is_empty() {
-                let version = tag.to_string();
-                info!("Git version (tag): {version}");
+        // Look for board-specific tag in format "board_name/version"
+        let tag_prefix = format!("{board_name}/");
```
LK, line 346: what do you think about making the board name comparison case-insensitive?

crates/pcb/tests/snapshots/simple__simple_workspace_release.snap
akhilles: We still try to place the release in workspace root `.pcb`. This specific test case has no `pcb.toml`, so the workspace root is detected as the parent dir of the target `.zen` file.

This a consequence of unifying the workspace root detection behavior. Prior to this change, `pcb release` had its own special workspace root detection which would do eval and then find common ancestor. But none of the other commands did this, so it was a bit inconsistent. Happy to revert this change though. It's not strictly related to this PR.

crates/pcb/src/release.rs
akhilles: I'm good with that.

crates/pcb/tests/snapshots/simple__simple_workspace_release.snap
LK: ah got it, this makes sense to me!

LK (approved):

crates/pcb/src/release.rs
akhilles: done

PR #105, opened 2025-08-23
Add warnings for unstable refs in `load()`, `Module()`

akhilles:
just warnings:
image
warnings + error:
image
`-Dwarnings`:
image


akhilles:
> Agreed. Rendering a span pointing to pcb.toml is a bit of work. For now, can we generate a single span-less warning per alias, and suppress all resolution paths that use that alias? and add spans as a follow-up

image

Added `pcb.toml` span, it wasn't too much work. I think it's fine to show it as a child diagnostic of the `.zen` callsite right? Just so you can track down where this alias is being used from. We can still batch to reduce noise and only show 1 instance of this.

TODO:
- Aggregation of similar warnings
- Filter out some warnings for LSP

akhilles:
Added diagnostics passes as a construct:
```rust
vec![
    Box::new(pcb_zen_core::FilterPass),                   // filter out <hidden> messages
    Box::new(pcb_zen_core::PromoteDeniedPass::new(deny)), // promote warnings to errors if "-Dwarnings"
    Box::new(pcb_zen_core::AggregatePass),                // aggregate similar warnings
    Box::new(pcb_zen_core::SortPass),                     // sort warnings before errors
    Box::new(pcb_zen::diagnostics::RenderPass),           // render diagnostics
]
```
Some existing functionality was refactored into diagnostics passes (e.g. filtering, rendering).

Also added aggregation as a diagnostics pass:
image

To facilitate this, I added type-erased error storage w/ downcasting to `Diagnostic`. This let's us carry additional typed metadata within `Diagnostic` that the passes can operate on.

I think the only thing missing is an LSP-specific filter pass

@LK lemme know if there's anything not addressed from your feedback

akhilles:
Add a diagnostics pass to filter out unstable ref warnings for transitive dependencies:

before:
image

after:
image

crates/pcb-zen-core/src/diagnostics.rs
```diff
@@ -0,0 +1,472 @@
...
+    /// Optional child diagnostic representing a nested error that occurred in a
+    /// downstream (e.g. loaded) module.  When present, this allows callers to
+    /// reconstruct a chain of diagnostics across module/evaluation boundaries
+    /// without needing to rely on parsing rendered strings.
+    pub child: Option<Box<Diagnostic>>,
+
+    /// Optional type-erased source error that can be downcast to specific error types
+    /// for detailed analysis while keeping the diagnostic struct generic-free.
+    /// Using Arc to preserve type information during clone operations.
+    pub source_error: Option<Arc<anyhow::Error>>,
```
LK, line 71: wonderful

LK (approved):
🙌

PR #115, opened 2025-08-25
Copy all tracked files to release src

akhilles:
Remove the filtering that I added, it's not needed. This can be further
simplified in the future by merging tracked_local_files into
path_to_spec.

Also, build after release in snapshot tests.

### DIFF
--- a/crates/pcb/src/release.rs
+++ b/crates/pcb/src/release.rs
@@ -19,6 +19,7 @@ use std::process::Command;
 use zip::{write::FileOptions, ZipWriter};
 
 use crate::bom::write_bom_json;
+use crate::build::evaluate_zen_file;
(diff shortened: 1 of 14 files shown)


crates/pcb/src/release.rs
```diff
@@ -406,45 +403,52 @@ fn copy_sources(info: &ReleaseInfo) -> Result<()> {
...
-                            "Failed to copy {} -> {}",
-                            path.display(),
-                            dest_path.display()
-                        )
-                    })?;
-                }
+        } else {
+            let Ok(rel) = path.strip_prefix(info.workspace.root()) else {
+                // local dep not in workspace
+                continue;
```
LK: Should we error in this case?

akhilles: yeah I think so. referring to paths outside the workspace should only be used during dev right?

akhilles: done

PR #117, opened 2025-08-26
Add `pcb tag`

akhilles:
Usage: `pcb tag -b [board_name] -v <version> [--push]`

image

Builtin validations:
- Board name validation - Board exists in workspace (if specified)
- Default board resolution - Has default board or single board (if none specified)
- Semantic version parsing - Valid semver format
- Tag uniqueness - Tag doesn't already exist
- Version ordering - New version is higher than existing versions for that board


crates/pcb/src/tag.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    // Check existing versions
+    let existing_tags = git::list_tags(&workspace_info.root, &format!("{board_name}/v*"))?;
+    for tag in existing_tags {
+        if let Some(v) = tag
+            .strip_prefix(&format!("{board_name}/v"))
+            .and_then(|s| Version::parse(s).ok())
+        {
+            if version <= v {
```
LK: We should probably allow non-monotonically-increasing version numbers in case we need/want to update a previous build of a board. I'm ok to show this as a warning/confirmation dialog?

crates/pcb/src/tag.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// Version to tag (must be valid semantic version)
+    #[arg(short = 'v', long, required = true)]
+    pub version: String,
+
+    /// Push tag to remote repository
+    #[arg(long)]
+    pub push: bool,
+
+    /// Optional path to start discovery from (defaults to current directory)
+    pub path: Option<String>,
```
LK, line 33: Thoughts on a `-f` to force-push/overwrite a release?

crates/pcb/src/tag.rs
akhilles: I'll just remove this check. It doesn't make much sense now that I think about it. No need to enforce a linear versioning model, enforcing uniqueness is fine.

crates/pcb/src/tag.rs
akhilles: Will add `-f` to skip confirmation for `--push`, but I'd be hesitant to overwrite tags on remote or bypass the other validations. I think using raw git commands provides appropriate friction when straying away from the happy path.

crates/pcb/src/tag.rs
akhilles: done

crates/pcb/src/tag.rs
akhilles: Added `-f`

LK: makes sense

PR #118, opened 2025-08-26
Add `Path()` with `allow_not_exist` support

akhilles:
The following will build fine even if the `build/Module` directory doesn't exist:
```
Layout(
    name = "Module",
    path = Path("build/Module", allow_not_exist=True),
)
```
and will be tracked + vendored properly.

We could hide this in `Layout()`, but the implementation gets a bit complicated, and it looks inconsistent with how we use `File()` elsewhere. I'm not sure it's worth it. If we make `Layout()` a language-level construct, it becomes trivial though.

### DIFF
--- a/crates/pcb/tests/path.rs
+++ b/crates/pcb/tests/path.rs
@@ -1,7 +1,8 @@
 #![cfg(not(target_os = "windows"))]
 
 use pcb_test_utils::assert_snapshot;
-use pcb_test_utils::sandbox::Sandbox;
+use pcb_test_utils::sandbox::{cargo_bin, Sandbox};
+use serde_json::Value;
 
 const SIMPLE_WORKSPACE_PCB_TOML: &str = r#"
 [workspace]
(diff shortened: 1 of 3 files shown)


crates/pcb/tests/path.rs
```diff
@@ -0,0 +1,203 @@
...
+    let staging_dir = json["staging_directory"]
+        .as_str()
+        .expect("Missing staging_directory in JSON");
+
+    // Snapshot the staging directory contents
+    assert_snapshot!("path_local_mixed_release", sb.snapshot_dir(staging_dir));
+}
+
+#[test]
+fn test_path_function_missing_without_allow() {
```
LK, line 178: Would it be simpler to have the `Path` tests in the `pcb-zen-core` tests?

LK (approved):

crates/pcb/tests/path.rs
akhilles: I can move the single file / local dependency tests easily, but the ones using git fixtures are a bit tricky. Also, I really want to test `vendor` + `release` behavior with `Path()` to validate tracking. So, testing through the CLI commands seemed useful in terms of coverage.

PR #124, opened 2025-08-27
Add `metadata()`, record ser/de support

akhilles:
Also added `push_metadata()`, `get_metadata()`, `list_metadata()`.

TODO:
- **Integrate with interfaces**
- Support nested records in metadata (imo, less of a priority)
- Improve metadata type validation (we type check simple types, but not records)
- Add variable arguments to `push_metadata()` (e.g. `push_metadata(container, a, b, c)`)
- Add `has_metadata()`

### DIFF
--- a/crates/pcb-zen-core/src/lang/metadata.rs
+++ b/crates/pcb-zen-core/src/lang/metadata.rs
@@ -1,0 +1,357 @@
+use std::collections::HashMap;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+
+use allocative::Allocative;
+use anyhow::anyhow;
+use serde_json;
+use starlark::any::ProvidesStaticType;
+use starlark::environment::GlobalsBuilder;
+use starlark::eval::Evaluator;
+use starlark::starlark_module;
+use starlark::starlark_simple_value;
+use starlark::values::list::AllocList;
+use starlark::values::starlark_value;
+use starlark::values::types::record::ty_record_type::TyRecordData;
+use starlark::values::types::record::{FrozenRecordType, RecordType};
+use starlark::values::{NoSerialize, StarlarkValue, Value, ValueLike};
+
+use crate::lang::evaluator_ext::EvaluatorExt;
+use crate::lang::input::InputValue;
+
+/// Helper to access metadata store with error handling
+fn with_metadata_store<T, F>(eval: &Evaluator, f: F) -> anyhow::Result<T>
+where
+    F: FnOnce(&MetadataStore) -> T,
+{
+    eval.context_value()
+        .ok_or_else(|| anyhow!("No evaluation context available for metadata operation"))?
+        .parent_context()
+        .with_metadata_store(f)
+}
+
+/// Helper to access mutable metadata store with error handling  
+fn with_metadata_store_mut<T, F>(eval: &Evaluator, f: F) -> anyhow::Result<anyhow::Result<T>>
+where
+    F: FnOnce(&mut MetadataStore) -> anyhow::Result<T>,
+{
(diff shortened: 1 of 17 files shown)


docs/pages/spec.mdx
```diff
@@ -392,6 +394,86 @@ if "SubmoduleName.ComponentName" in module:
...
+
+**Dimensionless Values**:
+- Dimensionless physical values can be cast to any unit type
+- Use `with_unit(None)` to convert any physical value to dimensionless
+- Dimensionless results are produced by unit cancellation (e.g., R×G = 1, V/V = 1)
+- Numeric literals are treated as dimensionless when mixed with physical values
+
+**String Representation**:
+Physical values maintain lossless string conversion, preserving original units and tolerances.
+
```
LK, line 490: can we also add spec for metadata here?

crates/pcb-zen-core/src/lang/metadata.rs
```diff
@@ -0,0 +1,371 @@
...
+        let input_values = with_metadata_store(eval, |store| {
+            store
+                .get_all(self.container.ref_id)
+                .cloned()
+                .unwrap_or_default()
+        })?;
+
+        let list_values = input_values
+            .iter()
+            .map(|input_value| input_value.to_value(eval, None))
```
LK, line 317: What happens if we have types where `expected_typ` is needed (e.g. enums, records, interfaces)?

crates/pcb-zen-core/src/lang/metadata.rs
```diff
@@ -0,0 +1,371 @@
...
+    }
+}
+
+#[starlark_value(type = "MetadataContainer")]
+impl<'v> StarlarkValue<'v> for MetadataContainer {
+    type Canonical = MetadataContainer;
+
+    fn get_attr(&self, attribute: &str, heap: &'v Heap) -> Option<Value<'v>> {
+        match attribute {
+            "push" => {
```
LK, line 127: this is very nice, I like this more than a bunch of global functions too

crates/pcb-zen-core/src/lang/metadata.rs
akhilles: I couldn't find a way to convert those types to values without explicitly requiring the type to be provided at the call site. There's no easy way to extract + carry the necessary type information in Rust.

Maybe we can have an optional type hint argument? e.g. `metadata.get(EnumTy)`. It's a bit awkward.

akhilles: I was able to get simple records and enums sorta working by hacking around in starlark-rust, but it was very brittle, and still failed on stuff like records containing enums.

LK: I'm ok to say they are not supported right now especially now with native units — what's the behavior if you try to construct a metadata container of type `enum(...)`?

akhilles: You can construct and push values into a metadata container with `enum()`. On `get()`, `list()`, it'll fail with "enum input requires expected_typ".

LK: Should we just disallow building a container with unsupported types for now?

docs/pages/spec.mdx
akhilles: done

crates/pcb-zen-core/src/lang/metadata.rs
akhilles: done

akhilles:
We have no pressing need to `metadata()`. Will revisit if/when there's a stronger case for it.

PR #129, opened 2025-08-30
[DIO-352] Add `TestBench()` support

akhilles:
To test:
```sh
cargo run -- build bench_test_cases/bmi270.zen      # should pass
cargo run -- build bench_test_cases/tc1/bench.zen   # should fail
```

To set up a bench, just pass in the module and some functions:
```python
TestBench(
    name="BMI270",
    module=BMI270,
    checks = [
        print_info_verbose,
        print_info,
        check_power_decoupling,
        check_bmi270_power_pins,
        check_communication_interfaces,
    ],
)
```

The information available to each check fn is the module under test, which has the following attrs:
- `nets`: net name -> list of ports
- `components`: component path -> `ComponentValue`
```
Nets and their connected ports:
  io_INT1_NET: ["BMI270.BMI270.INT1"]
  io_ois_MOSI: ["BMI270.BMI270.ASDX", "BMI270.BMI270.OSDO"]
  io_i2c_SDA: ["BMI270.BMI270.SDX"]
  io_INT2_NET: ["BMI270.BMI270.INT2"]
  io_gnd_GND: ["C1.C.P2", "C2.C.P2", "BMI270.BMI270.SDO", "BMI270.BMI270.GNDIO", "BMI270.BMI270.GND"]
  io_ois_CLK: ["BMI270.BMI270.ASCX"]
  io_i2c_SCL: ["BMI270.BMI270.SCX"]
  io_ois_CS: ["BMI270.BMI270.OCSB", "BMI270.BMI270.CSB"]
  io_vdd_VCC: ["C1.C.P1", "C2.C.P1", "BMI270.BMI270.VDDIO", "BMI270.BMI270.VDD"]
Components and their attributes:
  C2.C:
    Name: C
    Prefix: C
    Type: capacitor
    Capacitance: 100nF
    Properties: {"Capacitance": "100nF", "Package": "0402", "Type": "capacitor", "Value": "100nF"}
    Pins:
      P1: io_vdd_VCC
      P2: io_gnd_GND
  BMI270.BMI270:
    Name: BMI270
    Prefix: U
    MPN: BMI270
    Properties: {"mpn": "BMI270", "value": "BMI270"}
    Pins:
      ASCX: io_ois_CLK
      ASDX: io_ois_MOSI
      CSB: io_ois_CS
      GND: io_gnd_GND
      GNDIO: io_gnd_GND
      INT1: io_INT1_NET
      INT2: io_INT2_NET
      OCSB: io_ois_CS
      OSDO: io_ois_MOSI
      SCX: io_i2c_SCL
      SDO: io_gnd_GND
      SDX: io_i2c_SDA
      VDD: io_vdd_VCC
      VDDIO: io_vdd_VCC
  C1.C:
    Name: C
    Prefix: C
    Type: capacitor
    Capacitance: 1uF
    Properties: {"Capacitance": "1uF", "Package": "0402", "Value": "1uF", "Type": "capacitor"}
    Pins:
      P1: io_vdd_VCC
      P2: io_gnd_GND
```


crates/pcb-zen-core/src/graph/starlark.rs
```diff
@@ -0,0 +1,487 @@
...
+        let heap = eval.heap();
+
+        // Extract arguments from named parameters
+        let args_map = args.names_map()?;
+        let start = args_map.get(&heap.alloc_str("start")).copied();
+        let end = args_map.get(&heap.alloc_str("end")).copied();
+        let max_depth = args_map
+            .get(&heap.alloc_str("max_depth"))
+            .and_then(|v| v.unpack_i32())
+            .unwrap_or(10) as usize;
```
LK: do we detect cycles in the graph pathfinding algorithm? if so, wdyt about making the max_depth opt-in but by default leaving it uncapped? I'm a little worried about a situation where we might have a handful of components in series and we quickly blow through this limit and it won't be obvious why we can't find a path

LK (approved):
💯

crates/pcb-zen-core/src/graph/starlark.rs
akhilles: yes, we avoid cycles by tracking visited edges, vertices. will make max_depth opt in

PR #131, opened 2025-08-30
Hide netlist.json from layout output

LK:


akhilles:
Just curious, why hide this? I found it useful for debugging layout sync.

akhilles:
> You can always `pcb build --netlist` which should give you the same thing, but I don't think there's any reason we need to clutter up the layout folder with it

Ah didn't know about --netlist, that's perfect. I thought this just printed the path to the log file, didn't realize it would copy to layout dir.

hexdae (approved):

PR #135, opened 2025-09-04
Fix hierarchical net naming

akhilles:
Remove spurious net name deduplication.

```
Level1 = interface(
    net = Net("INNER"),
)

# Level 2: Interface that uses Level1
Level2 = interface(
    level1 = Level1(),
)

# Level 3: Interface that uses Level2
Level3 = interface(
    level2 = Level2(),
)

l1 = Level1()
print(l1.net.name)
l2 = Level2()
print(l2.level1.net.name)
l3 = Level3()
print(l3.level2.level1.net.name)

l1_named = Level1("L1")
print(l1_named.net.name)
l2_named = Level2("L2")
print(l2_named.level1.net.name)
l3_named = Level3("L3")
print(l3_named.level2.level1.net.name)
```

before:
```
INNER_3
INNER_4
INNER
L1_INNER
L2_LEVEL1_INNER
INNER
```

after:
```
INNER
LEVEL1_INNER
LEVEL2_LEVEL1_INNER
L1_INNER
L2_LEVEL1_INNER
L3_LEVEL2_LEVEL1_INNER
```

### DIFF
--- a/crates/pcb-zen-core/src/lang/interface.rs
+++ b/crates/pcb-zen-core/src/lang/interface.rs
@@ -112,6 +112,18 @@ fn interface_instance_factory<'v>(value: Value<'v>, _heap: &'v Heap) -> anyhow::
     }
 }
 
+/// Recursively unregister all nets within an interface instance
+fn unregister_interface_nets<'v>(interface: &InterfaceValue<'v>, ctx: &ContextValue) {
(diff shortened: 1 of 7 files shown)


akhilles:
> In this case, since `Level1` is a single-net interface, should we just call the net `L1`?

Yes, this is the next change I'm gonna make. Just wanted to have some predictable naming behavior before I made this change.

PR #145, opened 2025-09-08
Add basic BOM matching API 

akhilles:
See `crates/pcb-sch/src/bom.rs`. The main APIs are:

```rust
fn ungrouped_json()
fn grouped_json()
fn apply_bom_rule()
```

Added `--rule <path_to_bom_rules.json>` arg to `pcb bom` just to test out this feature. We can remove it once we have a discovery mechanism for these.

Also added `EvalConfig`, `pcb_zen::eval` to consolidate the eval code more, reduce duplication.


crates/pcb-sch/src/bom.rs
```diff
@@ -119,110 +158,184 @@ impl FromStr for Dielectric {
...
-pub fn generate_bom_entries(schematic: &mut Schematic) -> BTreeMap<String, BomEntry> {
-    let mut bom_entries = BTreeMap::new();
+// BOM Matching API
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum BomMatchingKey {
+    Mpn(String),
+    Resistor(ResistorMatchingKey),
+    Capacitor(CapacitorMatchingKey),
+    Path(String),
+    Designator(String),
```
LK: I'm a bit scared to do this because designators are very likely to shuffle around unpredictably, and this can easily make this BOM line attach to the wrong component. At least path changes are more explicit, and will usually just result in nothing matching this line — does `Path` alone give us enough of a fallback option?

crates/pcb-sch/src/bom.rs
```diff
@@ -119,110 +158,184 @@ impl FromStr for Dielectric {
...
-/// Generate ungrouped BOM entries from a schematic
-pub fn generate_bom_entries(schematic: &mut Schematic) -> BTreeMap<String, BomEntry> {
-    let mut bom_entries = BTreeMap::new();
+// BOM Matching API
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum BomMatchingKey {
+    Mpn(String),
+    Resistor(ResistorMatchingKey),
+    Capacitor(CapacitorMatchingKey),
+    Path(String),
```
LK: In the TS version of this I had `Path(Vec<String>)` so you can assign multiple elements to the same row, which might be convenient

crates/pcb-sch/src/bom.rs
```diff
@@ -119,110 +158,184 @@ impl FromStr for Dielectric {
...
 
-/// Generate ungrouped BOM entries from a schematic
-pub fn generate_bom_entries(schematic: &mut Schematic) -> BTreeMap<String, BomEntry> {
-    let mut bom_entries = BTreeMap::new();
+// BOM Matching API
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum BomMatchingKey {
+    Mpn(String),
+    Resistor(ResistorMatchingKey),
+    Capacitor(CapacitorMatchingKey),
```
LK: worth grouping these into a "Generic" enum that gets sub-specialized?

crates/pcb-zen-core/src/convert.rs
```diff
@@ -458,6 +457,50 @@ pub trait ToSchematic {
     fn to_schematic(&self) -> anyhow::Result<Schematic>;
 }
 
+/// Extract (value, tolerance, unit_name) from a Starlark unit record
+fn parse_unit_record<'v, V: ValueLike<'v>>(v: V) -> Option<pcb_sch::PhysicalValue> {
```
LK, line 461: 👍 eventually we should bring unit support up into the runtime, I think

crates/pcb-sch/src/bom.rs
```diff
@@ -1,81 +1,63 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
     value: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     description: Option<String>,
-    alternatives: Vec<String>,
+    #[serde(flatten)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    well_known_module: Option<WellKnownComponent>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    voltage: Option<PhysicalValue>,
```
LK: This might belong as a property within `WellKnownComponent`?

crates/pcb-sch/src/bom.rs
```diff
@@ -1,81 +1,63 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
     package: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     value: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     description: Option<String>,
-    alternatives: Vec<String>,
+    #[serde(flatten)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    well_known_module: Option<WellKnownComponent>,
```
LK: nit: `well_known_component`? or maybe something like `generic_data`?

crates/pcb-sch/src/bom.rs
```diff
@@ -1,81 +1,63 @@
...
-    pub mpn: Option<String>,
-    pub alternatives: Vec<String>,
-    pub package: Option<String>,
-    pub value: Option<String>,
-    pub description: Option<String>,
-    pub well_known_module: Option<WellKnownModule>,
-    pub voltage: Option<PhysicalValue>,
-    pub dnp: bool,
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+struct GroupedBomEntry {
```
LK, line 15: need to think about this one a little bit, but it might be nicer to always represent the "raw"/design/unmatched BOM as just a raw list of components with metadata, and then the "aggregation" happens by grouping components that match to the same match key. And then that match key should imply a set of potential distributor parts. We can do this either by saying each match rule corresponds to a single distributor part and aggregating over all of the match rules that apply to a specific component, or by letting each match rule point to a set of potential distributor part numbers?

Can talk about this more tomorrow

crates/pcb-zen-core/src/convert.rs
akhilles: yeah, this is a blocker for metadata as well

crates/pcb-sch/src/bom.rs
akhilles: yes, I want to reuse the capacitor, resistor structs we use in BomEntry as well.

crates/pcb-sch/src/bom.rs
akhilles: ack, I think path is fine. I end up resolving designator to path anyways, but avoiding designators in any persisted state makes sense.

crates/pcb-sch/src/bom.rs
akhilles: yup, will change to vec

crates/pcb-sch/src/bom.rs
akhilles: yeah, figured most generics might have a voltage rating, but it's a bit more subtle. will move into properties.

crates/pcb-sch/src/bom.rs
akhilles: ack, renaming to generic_data

crates/pcb-sch/src/bom.rs
akhilles: Agree the fundamentally we should always store and process BOM as list of individual components/paths. I think aggregation should just be a "view" of the bom, and we can have different views of the bom. `GroupedBomEntry` is basically just an ephemeral state.

I think it's worth having a single standard BOM data structure for design bom -> source bom -> order bom. It can get progressively enriched over time, but it's very convenient to have all the same "views" accessible at every stage.

crates/pcb-sch/src/bom.rs
```diff
@@ -161,14 +200,15 @@ pub enum BomMatchingKey {
...
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-#[serde(tag = "component_type")]
-pub enum GenericMatchingKey {
-    Resistor(Resistor),
-    Capacitor(Capacitor),
+pub struct GenericMatchingKey {
+    #[serde(flatten)]
+    pub component: GenericComponent,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub package: Option<String>,
```
LK: What do you think about making this required?

crates/pcb-sch/src/bom.rs
```diff
@@ -258,38 +298,36 @@ impl Bom {
...
+                        entry.offers.extend(rule.offers.clone());
                     }
                 }
             }
             BomMatchingKey::Mpn(mpn) => {
                 for entry in self.entries.values_mut() {
-                    if entry.mpn.as_ref() == Some(mpn) {
-                        entry.matched_part = Some(rule.value.clone());
+                    if entry.matches_mpn(mpn) {
+                        entry.offers.extend(rule.offers.clone());
```
LK: if I'm interpreting this right, each matching rule adds to the set of possible offers for each component, right? I think this is reasonable, we might want to also somehow store a mapping back that tells us which offer came from which rule, so when looking at an individual component it's easy to tell why an offer ended up there?

akhilles: added a matched_by

crates/pcb-sch/src/bom.rs
akhilles: done

PR #151, opened 2025-09-12
Parse position data, include it in netlist

akhilles:
I tried adding more structure to:

```
pub symbol_positions: HashMap<String, Position>,
```

E.g. using net ids, splitting this into multiple maps, but it was a wash in terms of reducing complexity. `savePositions()` ends up undoing all that work to get back to strings so it was a bit pointless. For now, I think working with strings and names is fine. Can revisit in the future.

### DIFF
--- a/crates/pcb-zen-core/src/convert.rs
+++ b/crates/pcb-zen-core/src/convert.rs
@@ -474,24 +474,19 @@ impl ModuleConverter {
                     rotation: pos.rotation,
                 };
 
-                // First check if this is a module/component position
-                if self.is_instance_position(key, instance_ref).is_some() {
-                    if let Some(instance) = self.schematic.instances.get_mut(instance_ref) {
-                        instance
-                            .component_positions
-                            .insert(key.to_owned(), position);
-                    }
-                } else if let Some(net_id) = self.find_net_id_for_key(key, module, instance_ref) {
-                    // This is a net position - can have multiple positions per net
-                    if let Some(instance) = self.schematic.instances.get_mut(instance_ref) {
-                        instance
-                            .net_positions
-                            .entry(net_id)
-                            .or_insert_with(Vec::new)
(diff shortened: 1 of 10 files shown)


crates/pcb-zen-core/src/lang/module.rs
```diff
@@ -51,7 +51,58 @@ use crate::lang::context::FrozenContextValue;
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Position({:.1}, {:.1}, {:.0})",
+            self.x, self.y, self.rotation
+        )
+    }
+}
+
+impl starlark::values::Freeze for Position {
```
LK, line 73: Does `starlark_simple_value!` not generate this?

crates/pcb-zen-core/src/convert.rs
```diff
@@ -451,6 +462,137 @@ impl ModuleConverter {
...
+                    instance.symbol_positions.insert(symbol_key, position);
+                }
+            }
+        }
+    }
+
+    fn is_instance_position(&self, key: &str, instance_ref: &InstanceRef) -> Option<()> {
+        // TODO: Remove this complex logic once we disallow "." in component names
+        // This tries all possible splits because component names can contain dots (e.g., "SMF6.0CA")
+        // Once we ban dots in names, we can go back to simple: key.split('.').try_fold(...)
```
LK, line 497: 🙏

crates/pcb-zen-core/src/lang/module.rs
akhilles: Nope, and apparently floats don't impl Freeze for some reason, so the derive macro didn't work either.

LK: 🙃

PR #152, opened 2025-09-13
Add `moved()` directive

akhilles:
See spec.mdx for actual semantics. Ended up going with a much simpler approach without globs/wildcard for now.

`pcb:sch` comments are updated on any save operation, we don't clean up old comments. `pcb layout` will always "repair" old footprints, groups.

What's expected of the user is to just add a `moved()` directive whenever they move something, and all the downstream artifacts will understand this.

### DIFF
--- a/docs/pages/spec.mdx
+++ b/docs/pages/spec.mdx
@@ -418,6 +418,52 @@ Properties accessible via the TestBench value:
 
 ## Built-in Functions
 
+### moved
+
+The `moved()` directive allows you to specify path remapping for refactoring support. When modules or components are moved or renamed, downstream artifacts (like layout files) may still reference the old paths. The `moved()` directive provides a mapping from old paths to new paths.
+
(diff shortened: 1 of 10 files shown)


docs/pages/spec.mdx
```diff
@@ -418,6 +418,41 @@ Properties accessible via the TestBench value:
 
 ## Built-in Functions
 
+### moved(old_path, new_path)
```
LK, line 421: wonderful

the expectation is that the .zen file should already be updated to new_path, right? Should we add some warnings/errors if either `old_path` still exists in the schematic or `new_path` does not exist?

LK (approved):

docs/pages/spec.mdx
akhilles: Yeah, warnings are a good idea. Added both.

PR #169, opened 2025-09-20
Make `load()`, `Module()` case-sensitive

akhilles:
```
Error: Case mismatch: expected 'PinSocket.zen', found 'pinsocket.zen'
    ╭─[ /Users/akhilles/src/diode/9mothers/modules/SocketPdm.zen:18:13 ]
    │
 18 │ PinSocket = Module("@stdlib/kicad/PinSocket.zen")
    │             ──────────────────┬──────────────────
    │                               ╰──────────────────── Case mismatch: expected 'PinSocket.zen', found 'pinsocket.zen'
    │
    ├─[ /Users/akhilles/src/diode/9mothers/modules/Uma16.zen:20:1 ]
    │
 20 │ ╭─▶ SocketPdm(
    ┆ ┆
 35 │ ├─▶ )
    │ │
    │ ╰─────── Error instantiating `SocketPdm`
────╯
```

### DIFF
--- a/crates/pcb-zen-core/src/lib.rs
+++ b/crates/pcb-zen-core/src/lib.rs
@@ -837,6 +837,11 @@ impl LoadResolver for CoreLoadResolver {
             ));
         }
 
+        // Validate filename case matches exactly on disk (but only if file exists or is required to exist)
+        if context.file_provider.exists(&resolved_path) {
+            validate_path_case(context.file_provider, &resolved_path)?;
+        }
+
         Ok(resolved_path)
     }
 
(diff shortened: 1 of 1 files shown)


crates/pcb-zen-core/src/lang/module.rs
```diff
@@ -1479,6 +1479,9 @@ where
             )));
         }
 
+        // Validate case
+        validate_path_case(file_provider.as_ref(), &resolved_path)?;
```
LK: Does it make sense to put this in the LoadResolver so it applies to other places we resolve load specs too? (symbols, layouts, etc)

akhilles: yup, def. made the change.

PR #180, opened 2025-09-24
Unified caching for `load()`, `Module()`

akhilles:
~7x speedup:

Before:
```
  Time (mean ± σ):      2.163 s ±  0.037 s    [User: 1.564 s, System: 0.585 s]
  Range (min … max):    2.110 s …  2.227 s    10 runs
```

After:
```
  Time (mean ± σ):     305.7 ms ±   2.1 ms    [User: 261.6 ms, System: 41.5 ms]
  Range (min … max):   303.2 ms … 309.3 ms    10 runs
```

### DIFF
--- a/crates/pcb-zen-core/src/lang/eval.rs
+++ b/crates/pcb-zen-core/src/lang/eval.rs
@@ -6,7 +6,6 @@ use std::{
 };
 
 use anyhow::anyhow;
-use starlark::typing::Interface;
 use starlark::{
     any::ProvidesStaticType,
     environment::{GlobalsBuilder, LibraryExtension},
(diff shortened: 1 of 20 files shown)


crates/pcb-zen-core/src/lang/eval.rs
```diff
@@ -1029,89 +974,19 @@ impl EvalContext {
...
-        // Safety check: ensure the path matches what we expect
-        if *actual_path != path {
-            panic!(
-                "Module path mismatch at index {actual_index}: expected '{path}', got '{actual_path}'"
-            );
-        }
-
-        self.get_codemap()
-            .map(|codemap| codemap.file_span(*span).resolve_span())
+        let span = ast
```
LK, line 980: beautiful

PR #188, opened 2025-09-28
Board-level configuration

akhilles:
Can setup constraints, design rules (incl net classes + colors), board stackup. `board_config.zen` will be moved into stdlib. Just included in this PR for easier review.

Will add netclass assignments in a follow-up PR.


crates/pcb-layout/src/scripts/update_layout_file.py
```diff
@@ -1309,12 +1309,199 @@ def get_footprint_by_uuid(self, uuid: str) -> Optional[FootprintInfo]:
 class SetupBoard(Step):
     """Set up the board for the sync process."""
 
-    def __init__(self, state: SyncState, board: pcbnew.BOARD):
+    def __init__(self, state: SyncState, board: pcbnew.BOARD, board_config_path: str = None, dont_sync_board_config: bool = False):
```
LK: uber-nit: can we make this `sync_board_config`, I get lost in the double-negatives

crates/pcb-layout/src/scripts/update_layout_file.py
```diff
@@ -1309,12 +1309,199 @@ def get_footprint_by_uuid(self, uuid: str) -> Optional[FootprintInfo]:
...
         self.state = state
         self.board = board
+        self.board_config_path = board_config_path
+        self.dont_sync_board_config = dont_sync_board_config
+
+    def _is_new_board(self) -> bool:
```
LK: nit: unused?

crates/pcb-layout/src/scripts/update_layout_file.py
```diff
@@ -1309,12 +1309,199 @@ def get_footprint_by_uuid(self, uuid: str) -> Optional[FootprintInfo]:
...
+        """Check if this is a newly created board by looking for existing footprints."""
+        return len(list(self.board.GetFootprints())) == 0
+
+    # Configuration table: (json_path, ds_attribute, display_name, [custom_setter])
+    CONFIG_MAPPINGS = [
```
LK, line 1319: this is very nice

crates/pcb-zen-core/src/lang/sexpr.rs
```diff
@@ -0,0 +1,501 @@
+use std::fmt;
```
LK: we have `pcb-sexpr` right, any reason to not use(/extend) that?

examples/board_config.zen
```diff
@@ -0,0 +1,379 @@
+# Copper constraints - all clearances and widths for copper features
```
LK, line 1: nit: remove / move to `examples`?

crates/pcb-layout/src/scripts/update_layout_file.py
akhilles: 👍  done

crates/pcb-layout/src/scripts/update_layout_file.py
akhilles: removed

examples/board_config.zen
akhilles: yup moved

crates/pcb-zen-core/src/lang/sexpr.rs
akhilles: Merged the changes into pcb-sexpr. Originally this was a much smaller file just doing sexpr gen for stackup.

Issue #67, opened 2025-08-14
blinky.zen: Layout generation failed


Majestychen:
After executing the "pcb layout blinky.zen" command, the following error occurred. I don't know how to troubleshoot the error. I hope there can be a more detailed error report.

pcb layout blinky.zen
✗ blinky.zen: Layout generation failed
  Error: PCB generation error: Failed to create layout file: F:\download\pcb-x86_64-pc-windows-msvc\layout/layout.kicad_pcb
Error: Layout generation failed with errors

LK:
Thanks for the report, let me investigate; I suspect it's a bug with how we handle Windows paths.

GastonDonikian:
Hey, adding to this PR, very similar issue on Mac. 

Followed along the instructions from the README,

`pcb build blinky.zen` // built with warnings

`pcb layout blinky.zen` // failed with following error message:

✗ blinky.zen: Layout generation failed
  Error: PCB generation error: Failed to create layout file: layout.kicad_pcb
Error: Layout generation failed with errors

Warnings seem to be unrelated: They are about using an unstable reference for @stdlib.

akhilles:
I just merged a PR (#212) that improves the error messages for `pcb layout`. Could you try running `pcb layout blinky.zen` again? You should now get a more detailed error message that explains why the layout file creation is failing. I suspect the issue might be related to your KiCad installation or how the tool is trying to invoke KiCad. The improved error output should help us narrow down the root cause.

PR #212, opened 2025-10-05
Add better error messages, update `blinky.zen` (and add it to CI)

akhilles:
This can help debug https://github.com/diodeinc/pcb/issues/67.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -108,7 +108,7 @@ jobs:
         run: cargo test --verbose -p pcb -- --ignored
 
       - name: Build examples
-        run: cargo run -- build examples/blinky.zen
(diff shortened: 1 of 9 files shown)


.github/workflows/test.yml
```diff
@@ -107,6 +107,9 @@ jobs:
           RUST_BACKTRACE: full
         run: cargo test --verbose -p pcb -- --ignored
 
+      - name: Build examples
```
LK, line 110: Should we build everything in the examples folder?

LK (approved):

.github/workflows/test.yml
akhilles: yes definitely, done

PR #218, opened 2025-10-07
Add GLB export w/ gltfpacking

akhilles:
Also separate 3d model exports into multiple tasks. Runs `gltfpack` if available.

### DIFF
--- a/crates/pcb/src/release.rs
+++ b/crates/pcb/src/release.rs
@@ -1170,27 +1170,13 @@ fn generate_glb_model(info: &ReleaseInfo) -> Result<()> {
         }
     }
 
-    // Optimize GLB file with gltfpack if available
-    let optimize_result = Command::new("gltfpack")
-        .arg("-i")
-        .arg(&glb_path)
-        .arg("-cc")
-        .arg("-o")
-        .arg(&glb_path)
-        .output();
-
-    match optimize_result {
-        Ok(output) if output.status.success() => {
+    // Optimize GLB file with gltfpack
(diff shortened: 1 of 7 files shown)


crates/pcb/src/release.rs
```diff
@@ -1112,6 +1129,84 @@ fn generate_3d_models(info: &ReleaseInfo) -> Result<()> {
...
+    if let Err(e) = glb_result {
+        if glb_path.exists() {
+            warn!("KiCad CLI reported error but GLB file was created: {e}");
+        } else {
+            return Err(e).context("Failed to generate GLB model");
+        }
+    }
+
+    // Optimize GLB file with gltfpack if available
+    let optimize_result = Command::new("gltfpack")
```
LK: expecting gltfpack in CI and dev machines is a little annoying, how about we don't compress the glb in the release and we can instead do it as a post-processing step on the web when we receive the release? or alternatively embed `gltfpack` into pcb?

akhilles: currently, its just opportunistic. if `gltfpack` is missing, it doesn't optimize. the motivation was to create smaller releases so we don't make the load times worse. I can look into embedding

akhilles: also, I guess if we're unpacking the release, its less of an issue if the uploaded .zip file is bug.

akhilles: switched to gltfpack-sys

LK (approved):
beautiful, thanks!

PR #221, opened 2025-10-09
Typed nets w/ metadata

akhilles:
Adds typed nets:
```starlark
Net = builtin.net("Net")
Power = builtin.net("Power")
Ground = builtin.net("Ground")
```

Fully backwards compatible. The motivation for this is to add net metadata (voltage, impedance) at initialization-time for single-net interfaces, which are just special types of nets. Also adds support for Net casting. e.g.:

```starlark
net = Net(Power("VCC"))
# instead of:
net = Power("VCC").NET
```


crates/pcb-zen-core/src/lang/builtin.rs
```diff
@@ -195,42 +196,36 @@ fn builtin_methods(methods: &mut MethodsBuilder) {
         Ok(PhysicalValueType::new(unit.into()))
     }
 
+    fn net<'v>(
```
LK: would `.net_type` or something be more accurate?

crates/pcb-zen-core/src/lang/module.rs
```diff
@@ -1015,24 +1015,46 @@ fn validate_type<'v>(
...
-                return Ok(());
-            }
-        }
-        "InterfaceFactory" => {
-            if value.downcast_ref::<InterfaceValue>().is_some()
-                || value.downcast_ref::<FrozenInterfaceValue>().is_some()
-            {
+        if let Some(actual_type_name) = actual_type_name {
+            // Only allow exact type match - conversion will be handled by try_net_conversion
+            if expected_type_name == actual_type_name {
```
LK, line 1038: If reading this correctly, all `builtin.net("XYZ")` with the same `"XYZ"` are compatible net types, right?

akhilles: yes, this means "Power" and "Net" are considered incompatible here. but then there's `try_net_conversion` as a fallback.

PR #223, opened 2025-10-09
Replace InputValue with deep copy and native json ser/de

akhilles:


crates/pcb-zen-core/src/lang/eval.rs
```diff
@@ -121,6 +133,52 @@ pub(crate) fn copy_value<'dst>(v: Value<'_>, dst: &'dst Heap) -> anyhow::Result<
...
+        } else {
+            &repr
+        };
+
+        // Return variant as string - try_enum_conversion() will convert back to EnumValue
+        return Ok(dst.alloc_str(variant).to_value());
+    }
+
+    // Handle Record from external starlark crate
+    // Record doesn't implement DeepCopyToHeap. We convert to dict {field: value}
```
LK: discussed offline but I think we can implement DeepCopyToHeap for records, enums

akhilles: donezo

PR #235, opened 2025-10-16
Multi-unit symbol position support

LK:


crates/pcb-zen-core/src/convert.rs
```diff
@@ -701,8 +701,13 @@ impl ModuleConverter {
     }
 
     fn is_instance_position(&self, key: &str, instance_ref: &InstanceRef) -> Option<()> {
+        // Strip @U suffix from the key if present (for multi-unit symbols)
+        // e.g., "U1.OPEN_Q_6490CS@U1" -> "U1.OPEN_Q_6490CS"
+        let key_without_unit = key.split('@').next().unwrap_or(key);
```
LK, line 706: I guess we should disallow `@` in component names as part of this too, which I don't think should break anything

akhilles: yeah, could you add this to `validate_identifier_name()` in `lang/validation.rs`?

PR #240, opened 2025-10-21
Add `pcb mcp`

akhilles:


crates/pcb/src/mcp.rs
```diff
@@ -0,0 +1,32 @@
...
+pub struct McpArgs {}
+
+pub fn execute(_args: McpArgs) -> Result<()> {
+    let mut tools = vec![];
+
+    #[cfg(feature = "api")]
+    tools.extend(pcb_diode_api::mcp::tools());
+
+    // Point to public Zener documentation (client fetches directly)
+    let resources = vec![ResourceInfo {
```
LK, line 15: The reason we have this as a resource and a tool is that amp doesn't support resources?

akhilles: yep

PR #241, opened 2025-10-21
Add component modifiers + interior mutability

akhilles:


crates/pcb-zen-core/tests/component.rs
```diff
@@ -216,7 +213,413 @@ snapshot_eval!(component_inherits_reference_prefix, {
...
+
+snapshot_eval!(component_modifier_basic, {
+    "test.zen" => r#"
+        # Test component modifier
+        def assign_part(component):
+            if hasattr(component, "resistance"):
+                component.mpn = "ASSIGNED_MPN"
+                component.manufacturer = "ACME"
+
+        builtin.add_component_modifier(assign_part)
```
LK, line 235: I'm seeing a behavior where the modifier is only applied to components that are initialized after the modified is attached, is that expected? (meaning if I move this line below the `Component(...)` instantiation below, then it won't apply to that component)

akhilles: This was intentional, but I think worth revisiting. It was mainly a easy of implementation thing. I wanted to hook into `Component()`. But it should be easy enough to do post module eval, but pre-freeze.

LK: If we're using this for BOM rule resolution as part of `Board` then will probably need to apply retroactively as part of module finalization

akhilles: Fixed, should be order independent now.

LK (approved):

PR #245, opened 2025-10-21
Add HTML fab drawing to release

akhilles:
Writing to manufacturing/fab_drawing.html.

### DIFF
--- a/crates/pcb-layout/src/fab_drawing.rs
+++ b/crates/pcb-layout/src/fab_drawing.rs
@@ -170,8 +170,8 @@ pub fn generate_html(board_config: &BoardConfig) -> String {
 
     let copper_count = copper_thicknesses.len();
     assert!(
-        copper_count > 2,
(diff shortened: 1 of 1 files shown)


crates/pcb-layout/src/fab_drawing.rs
```diff
@@ -0,0 +1,318 @@
...
+            }
+        })
+        .collect();
+
+    let copper_count = copper_thicknesses.len();
+    let total_thickness = stackup
+        .thickness
+        .unwrap_or_else(|| layers.iter().map(|l| l.thickness()).sum());
+    let outer_oz = copper_thicknesses.first().map(|t| t / 0.035);
+    let inner_oz = if copper_count > 2 {
```
LK: Maybe worth throwing if we find something weird here? (similarly for rounding copper weight)

akhilles: done

PR #251, opened 2025-10-24
Store module tree in eval context instead of module starlark value

akhilles:
Has a ~20% perf improvement on complex boards
```
# before
> hyperfine "pcb build boards"
Benchmark 1: pcb build boards
  Time (mean ± σ):      5.146 s ±  0.620 s    [User: 3.282 s, System: 1.377 s]
  Range (min … max):    4.852 s …  6.859 s    10 runs

# after
> hyperfine "pcb build boards"
Benchmark 1: pcb build boards
  Time (mean ± σ):      3.971 s ±  0.053 s    [User: 2.511 s, System: 1.030 s]
  Range (min … max):    3.902 s …  4.086 s    10 runs
```

### DIFF
--- a/crates/pcb-zen-core/src/lang/eval.rs
+++ b/crates/pcb-zen-core/src/lang/eval.rs
@@ -1337,7 +1337,6 @@ impl EvalContext {
     }
 
     /// Process a pending child after the parent module has been frozen
-    /// Returns the child's FrozenModule to keep its heap alive
     fn process_pending_child(
         mut self,
         pending: FrozenPendingChild,


crates/pcb-zen-core/src/lang/eval.rs
```diff
@@ -1235,6 +1335,93 @@ impl EvalContext {
             Err(diagnostic.into())
         }
     }
+
+    /// Process a pending child after the parent module has been frozen
+    /// Returns the child's FrozenModule to keep its heap alive
```
LK: nit: I think this is no longer true, we just add it directly to the tree right?

PR #320, opened 2025-12-04
Fix Windows path separator in workspace URL generation

akhilles:
This was causing Windows CI to fail on v2 registry: https://github.com/diodeinc/registry/actions/runs/19919791686/job/57106022057.


crates/pcb-zen/src/workspace.rs
```diff
@@ -15,6 +15,8 @@ use std::collections::{BTreeMap, HashMap, HashSet};
 use std::path::{Path, PathBuf};
 use walkdir::WalkDir;
 
+use path_slash::PathExt as _;
```
LK, line 18: TIL

LK (approved):

PR #354, opened 2025-12-15
[FEAT] style warnings

hexdae:
# Description

Add code style advice diagnostics and promote them to warnings in the LSP to enforce a more cohesive coding standard for the models

### DIFF
--- a/crates/pcb-zen/tests/module_loading.rs
+++ b/crates/pcb-zen/tests/module_loading.rs
@@ -95,8 +95,8 @@ local = "./modules"
         "modules/MyModule.zen",
         r#"
 # A simple module
-input = io("input", Net)
-output = io("output", Net)
+INPUT = io("INPUT", Net)
+OUTPUT = io("OUTPUT", Net)
 "#,
     );
 
@@ -108,8 +108,8 @@ MyModule = Module("@local/MyModule.zen")
 
 MyModule(
     name = "M1",
-    input = Net("IN"),
-    output = Net("OUT"),
+    INPUT = Net("IN"),
+    OUTPUT = Net("OUT"),
 )
 "#,
     );
@@ -354,8 +354,8 @@ fn module_relative_from_subdir() {
         "modules/MyModule.zen",
         r#"
 # A simple module
-input = io("input", Net)
-output = io("output", Net)
+INPUT = io("INPUT", Net)
+OUTPUT = io("OUTPUT", Net)
 "#,
     );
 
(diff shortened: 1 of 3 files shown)


crates/pcb-zen-core/src/lang/naming.rs
```diff
@@ -0,0 +1,345 @@
...
+        } else {
+            return false;
+        }
+    }
+
+    true
+}
+
+/// Convert a name to UPPERCASE convention.
+pub fn to_uppercase(name: &str) -> String {
```
akhilles, line 55: Not sure this is needed.

crates/pcb-zen-core/src/lang/naming.rs
```diff
@@ -0,0 +1,345 @@
...
+/// Check if a name follows snake_case convention.
+///
+/// snake_case names:
+/// - Consist only of lowercase letters, digits, and underscores
+/// - Must start with a lowercase letter
+/// - Cannot have consecutive underscores
+/// - Cannot end with an underscore
+///
+/// Examples: `enable_debug`, `num_channels`, `output_voltage`
+pub fn is_snake_case(name: &str) -> bool {
```
akhilles, line 47: IMO, simpler to just run `to_snake_case` and check for equivalence. This way you don't have to keep the two consistent.

crates/pcb-zen-core/src/lang/naming.rs
```diff
@@ -0,0 +1,345 @@
...
+    true
+}
+
+/// Convert a name to UPPERCASE convention.
+pub fn to_uppercase(name: &str) -> String {
+    name.to_ascii_uppercase()
+}
+
+/// Convert a name to snake_case convention.
+pub fn to_snake_case(name: &str) -> String {
```
akhilles, line 60: Suggest using https://docs.rs/heck/0.5.0/heck/trait.ToSnekCase.html

akhilles (commented):
Suggest using https://docs.rs/heck. it has `ToSnekCase`, `ToShoutySnekCase`. We already depend on this crate.

crates/pcb-zen-core/src/lang/naming.rs
hexdae: done

crates/pcb-zen-core/src/lang/naming.rs
hexdae: done

crates/pcb-zen-core/src/lang/naming.rs
hexdae: we might want to inject things like promoting `~` to `N` in the future, so I would rather have this here to do that in our naming conventions

PR #374, opened 2025-12-20
Format the auto-generated component .zen file

akhilles:
Make manufacturer, alternatives configureable.


crates/pcb-diode-api/templates/component.zen.jinja
```diff
@@ -5,10 +5,12 @@
 {{ description }}
 {%- endif %}
 
-Auto-generated using `pcb search`. Do not modify.
+Auto-generated using `pcb search`.
```
LK, line 8: 🙌

PR #380, opened 2026-01-03
Add `pcb search --registry` TUI

akhilles:
Screenshot 2026-01-02 at 8 15 46 PM

### DIFF
--- a/crates/pcb-diode-api/src/registry/tui/app.rs
+++ b/crates/pcb-diode-api/src/registry/tui/app.rs
@@ -511,7 +511,14 @@ impl App {
                     if let Ok(client) = RegistryClient::open() {
                         self.parts_count = client.count().unwrap_or(0);
                     }
-                    self.last_query.clear();
+                    // Trigger re-search with updated DB
+                    self.query_counter += 1;
+                    self.last_query = self.current_query();
+                    let _ = self.query_tx.send(SearchQuery {
+                        id: self.query_counter,
+                        text: self.last_query.clone(),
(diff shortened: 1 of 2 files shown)


LK (approved):
This is awesome. A few notes, can address in a separate PR if you prefer:

- can we gracefully fall back on just local index data (trigram matching?) when without internet?
- There is some input lag when typing on my measly M1, assuming some of the index matching logic is running in the render loop?
- I couldn't figure out how to use the slash commands, since I can only trigger the / menu when the search query is empty. and maybe the default copy behavior should copy the `Module(...)` load command since I assume that'll be the primary use case?

akhilles:
> * I couldn't figure out how to use the slash commands, since I can only trigger the / menu when the search query is empty. and maybe the default copy behavior should copy the `Module(...)` load command since I assume that'll be the primary use case?

I decided to copy Amp. so there's only command palette (Ctrl + O), no slash commands. but `/` on empty search also launches the command palette. wdyt about making the default copy behavior copying the fully qualified url? e.g. "github.com/diodeinc/registry/reference/TPS56A37RPAR" so that it can easily be used `Module()` or pasted in pcb.toml.

akhilles:
> There is some input lag when typing on my measly M1, assuming some of the index matching logic is running in the render loop?

Made some tweaks to the debouncing and made the package details fetching async, lmk if this is any better.

> can we gracefully fall back on just local index data (trigram matching?) when without internet?

yup, will address in a future PR.

PR #436, opened 2026-01-16
Make `pcb build` fast again

LK:
> [!NOTE]
> Introduces automated build-performance measurement and substantial runtime optimizations across core, EDA, and LSP.
> 
> - CI: New `Build Performance` workflow runs `bin/pcb-build-perf` on demo/arduino workspaces, compares base vs head, and comments significant changes on PRs
> - Tooling: Adds `bin/pcb-build-perf` (timed runs with stats/CI95) and `bin/analyze-profile` (Firefox Profiler analysis)
> - Runtime perf/logging: Adds `mimalloc`, `tracing`, `tracing-subscriber`, `tracing-chrome`; optional `rayon` for parallelism
> - pcb-eda: Implements lazy KiCad symbol-library scanning/on-demand parsing with caching and cycle handling; exposes `KicadPin` and getters; adds `regex`
> - pcb-zen-core: Introduces `EvalSession`/`EvalContextConfig`, thread-safe caches (RwLock), parallel child instantiation (rayon), global atomic net IDs, tracing spans, updated module_tree accessors, stricter path-case validation, and symbol-library cache/invalidation
> - pcb-sch: `Schematic::to_json` now emits canonical JSON (RFC 8785)
> - LSP: Handles watched file changes, coalesces rapid edits, optional netlist push (`zener/netlistUpdated`), and context hooks for open/close/changed
> - Misc: Pre-commit `cargo fmt` no longer runs with `--check`; tests/utilities updated to sanitize dynamic IDs
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 883eb0e8d23cc7e511d892d24df09bd3e32f3e2f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/pcb-zen/src/cache_index.rs
+++ b/crates/pcb-zen/src/cache_index.rs
@@ -2,11 +2,12 @@
 
 use anyhow::{Context, Result};
 use pcb_zen_core::config::Lockfile;
-use rusqlite::{params, Connection, OptionalExtension};
+use r2d2::{Pool, PooledConnection};
+use r2d2_sqlite::SqliteConnectionManager;
+use rusqlite::{params, OptionalExtension};
 use semver::Version;
 use std::collections::BTreeMap;
 use std::path::PathBuf;
-use std::sync::Mutex;
 
 use crate::git;
 use crate::tags;
@@ -16,7 +17,7 @@ use crate::tags;
 const SCHEMA_VERSION: i32 = 3;
 
 pub struct CacheIndex {
-    conn: Mutex<Connection>,
+    pool: Pool<SqliteConnectionManager>,
 }
 
 impl CacheIndex {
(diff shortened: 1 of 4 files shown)


hexdae:
Generated Image January 17, 2026 - 3_38PM

crates/pcb/Cargo.toml
```diff
@@ -9,8 +9,12 @@ authors = { workspace = true }
 default-run = "pcb"
 
 [features]
-default = ["api"]
+default = ["api", "mimalloc"]
 api = ["dep:pcb-diode-api"]
+# Custom allocators (mutually exclusive) - only one should be enabled at a time
+mimalloc = ["dep:mimalloc"]
+jemalloc = ["dep:tikv-jemallocator"]
```
akhilles: Should we drop the other allocators? I assume you wanted to test these, but mimalloc is a good default w/ stdlib fallback. Generally, cargo features should only be additive. I don't think it matters much in this case, but I always prefer fewer feature flags.

crates/pcb-zen/src/cache_index.rs
```diff
@@ -15,7 +16,7 @@ use crate::tags;
 const SCHEMA_VERSION: i32 = 3;
 
 pub struct CacheIndex {
-    conn: Connection,
+    conn: Mutex<Connection>,
```
akhilles: Could we use https://docs.rs/r2d2_sqlite here instead? In sqlite WAL mode, we can have multiple connections open.

crates/pcb/Cargo.toml
LK: yessir done

crates/pcb-zen/src/cache_index.rs
LK: done

PR #476, opened 2026-01-29
Repurpose Net Kind

Hugo-Villagrana:
Repurpose Kind to carry the Zener net type name.

Kind was previously unused. This change uses it to store the net type name as provided by Zener (e.g. "Net", "Power", "NotConnected").

The field is intentionally modeled as a `String` for now, since the set of net types is open-ended and user-defined. We can introduce a stronger abstraction later once the mechanics are clearer.

---

> [!NOTE]
> **Medium Risk**
> Touches core netlist schema/serialization and conversion logic, so downstream consumers/tests may break if they assumed the old `NetKind` enum or specific `kind` values.
> 
> **Overview**
> **Repurposes net kind to carry the Zener net type name.** `pcb-sch::Net.kind` is changed from a closed `NetKind` enum to an open-ended `String`, removing the old Ground/Power/Normal classification and updating constructors/usages accordingly.
> 
> `pcb-zen-core` now preserves a net’s *original* Starlark type (even if it was later coerced to `Net`) via a new `original_type_name` field on `NetValue`, and the schematic converter exports that value into `Net.kind`. Net assembly is refactored around a single `NetInfo` accumulator (name/ports/properties/type), and snapshots/tests are updated to reflect the new serialized `kind` values (e.g. `"Net"`) and added `original_type_name` metadata.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 95af66b3687a97cd81e0610d6b5f971c1bb206e7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/pcb-zen-core/src/lang/net.rs
+++ b/crates/pcb-zen-core/src/lang/net.rs
@@ -214,8 +214,8 @@ impl<'v, V: ValueLike<'v>> NetValueGen<V> {
     /// any Power/Ground -> Net coercion, or the current type_name if never coerced.
     pub fn logical_type_name(&self) -> &str {
         if self.original_type_name.is_empty() {
-            return &self.type_name
-        } 
+            return &self.type_name;
+        }
 
         &self.original_type_name
     }


crates/pcb-zen-core/src/lang/net.rs
```diff
@@ -98,8 +98,10 @@ pub struct NetValueGen<V> {
     pub(crate) name: String,
     /// The name originally requested before deduplication
     pub original_name: Option<String>,
-    /// The type name (e.g., "Net", "Power", "Ground")
+    /// The final type name after type conversion
     pub(crate) type_name: String,
+    /// Original type name before any type coercion (e.g. Power -> Net)
+    pub original_type_name: Option<String>,
```
akhilles: Don't think this needs to be optional, will simplify some stuff downstream.

Hugo-Villagrana: Sounds good will change, thanks!

PR #588, opened 2026-02-28
[SIM] more powerful tools

dioderobot:
# Description

Make `pcb simulate` a first class citizen with LSP support and reload on save.

---

> [!NOTE]
> **Medium Risk**
> Adds a new simulation execution path that shells out to `ngspice` (with timeouts) and integrates save-time diagnostics into the LSP, which could affect developer workflows and error reporting. Risk is moderate due to new external-process handling and changes to eval/LSP caching behavior.
> 
> **Overview**
> Makes `pcb simulate` a first-class simulation command: it now discovers `.zen` files from a workspace/directory, runs `ngspice` directly, and supports `--netlist` (stdout) and `--output` (write `.cir` then exit) for single-file use.
> 
> Introduces inline sim setup via `builtin.set_sim_setup(file=...|content=...)`, persists the setup (and call-site span) on the schematic, and extends netlist generation to append the setup and **fail fast** if any non-DNP components lack a `SpiceModel`.
> 
> Adds an `ngspice` runner (`pcb-sim/ngspice.rs`) with platform-specific path discovery and a default timeout, plus a generic timeout option in `pcb-command-runner`. The LSP now runs additional **save-time** diagnostics: on save it reuses cached eval output, generates a `.cir`, checks/runs `ngspice`, and publishes simulation errors/info pinned to the `set_sim_setup()` call site.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 225b6c8726371da7b2a3a4db4e36efc43b04c2fd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

<a href="https://app.devin.ai/review/diodeinc/pcb/pull/588" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/pcb/src/sim.rs
+++ b/crates/pcb/src/sim.rs
@@ -21,8 +21,8 @@ pub struct SimArgs {
     #[arg(long, value_hint = clap::ValueHint::FilePath)]
     pub setup: Option<PathBuf>,
 
-    /// Write .cir to a specific file (ngspice still runs on it). Only valid when simulating a single file.
-    #[arg(short, long, value_hint = clap::ValueHint::FilePath)]
+    /// Write .cir to a file and exit (ngspice is not run). Only valid when simulating a single file.
+    #[arg(short, long, value_hint = clap::ValueHint::FilePath, conflicts_with = "netlist")]
     pub output: Option<PathBuf>,
 
     /// Disable network access (offline mode) - only use vendored dependencies
@@ -94,28 +94,28 @@ fn simulate_one(
         writeln!(buf, "{setup}")?;
     }
 
-    // --netlist or `-o -`: print to stdout and return
-    let print_to_stdout = args.netlist || args.output.as_deref() == Some(std::path::Path::new("-"));
-    if print_to_stdout {
+    // --netlist: print to stdout and return (skip ngspice)
+    if args.netlist {
         std::io::stdout().write_all(&buf)?;
+        return Ok(true);
+    }
+
+    // --output: write .cir to file and return (skip ngspice)
+    if let Some(output_path) = &args.output {
+        File::create(output_path)?.write_all(&buf)?;
         return Ok(true);
     }
 
     // Write .cir next to the zen file so ngspice resolves relative paths correctly
     let zen_dir = zen_path.parent().unwrap_or(std::path::Path::new("."));
-    let cir_path: Box<dyn AsRef<std::path::Path>> = if let Some(output_path) = &args.output {
-        File::create(output_path)?.write_all(&buf)?;
-        Box::new(output_path.clone())
-    } else {
-        let mut tmp = tempfile::Builder::new()
-            .suffix(".cir")
-            .tempfile_in(zen_dir)?;
-        tmp.write_all(&buf)?;
-        tmp.flush()?;
-        Box::new(tmp.into_temp_path())
-    };
+    let mut tmp = tempfile::Builder::new()
+        .suffix(".cir")
+        .tempfile_in(zen_dir)?;
+    tmp.write_all(&buf)?;
+    tmp.flush()?;
+    let cir_path = tmp.into_temp_path();
 
-    let result = run_ngspice_captured((*cir_path).as_ref(), zen_dir)?;
+    let result = run_ngspice_captured(cir_path.as_ref(), zen_dir)?;
 
     if result.success {
         if args.verbose {
@@ -150,8 +150,10 @@ pub fn execute(args: SimArgs) -> Result<()> {
     }
 
     // Directory / workspace mode — behave like `pcb build`
-    if args.setup.is_some() || args.output.is_some() {
-        anyhow::bail!("--setup and --output are only supported when simulating a single file");
+    if args.setup.is_some() || args.output.is_some() || args.netlist {
+        anyhow::bail!(
+            "--setup, --output, and --netlist are only supported when simulating a single file"
+        );
     }
 
     let resolution_result =


crates/pcb-zen/src/lsp/mod.rs
```diff
@@ -567,6 +567,104 @@ impl LspContext for LspEvalContext {
...
     }
 
+    fn on_save_diagnostics(&self, uri: &LspUrl) -> Vec<lsp_types::Diagnostic> {
+        let path = match uri {
+            LspUrl::File(p) => p,
+            _ => return vec![],
+        };
+
+        // Full eval to get schematic (same pattern as evaluate_with_inputs)
+        let config = self.config_for(path);
```
LK: are we doing a second eval here bc you want to use the version of the files saved on disk instead of open buffers?

crates/pcb-zen/src/lsp/mod.rs
```diff
@@ -567,6 +567,104 @@ impl LspContext for LspEvalContext {
...
+        let zen_dir = path.parent().unwrap_or(std::path::Path::new("."));
+        let mut tmp = match tempfile::Builder::new().suffix(".cir").tempfile_in(zen_dir) {
+            Ok(t) => t,
+            Err(_) => return vec![],
+        };
+        if std::io::Write::write_all(&mut tmp, &buf).is_err() {
+            return vec![];
+        }
+
+        match pcb_sim::run_ngspice_captured(tmp.path(), zen_dir) {
```
LK, line 656: can we add a timeout to this (or better yet run it out of band, but might be hard because I think LSP isn't async) — long sims will lock up the LSP

CHANGELOG.md
```diff
@@ -8,6 +8,21 @@ and this project adheres to Semantic Versioning (https://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- `pcb simulate` (renamed from `pcb sim`) runs SPICE simulations directly via ngspice, replacing the previous shell-script workflow.
```
LK: also as wonderful as this is maybe condense the changelog entry a bit

dioderobot: will do

crates/pcb-zen/src/lsp/mod.rs
dioderobot: The second eval is because `validate_and_collect` runs `parse_and_analyze_file` (produces AST + diagnostics), but the sim path needs a full schematic via `eval()` → `to_schematic()`. It does use the open buffer contents via `get_load_contents` (not disk). Could cache the schematic to avoid the double eval but that would be a bigger refactor.

crates/pcb-zen/src/lsp/mod.rs
dioderobot: Done — added a 5s timeout via `CommandRunner::timeout()`. ngspice gets killed if it exceeds it, and the LSP returns a timeout diagnostic.

CHANGELOG.md
dioderobot: Condensed into a single bullet point.

crates/pcb-sim/src/ngspice.rs
```diff
@@ -0,0 +1,125 @@
...
+/// Result of running ngspice in captured mode (for LSP/programmatic use).
+pub struct SimulationResult {
+    pub success: bool,
+    /// Plain text output (ANSI stripped).
+    pub output: String,
+}
+
+/// Check if ngspice is installed and return a helpful error if not.
+pub fn check_ngspice_installed() -> Result<String> {
+    let path = ngspice_path();
```
akhilles, line 73: Is the hard-coded path search needed? Can we just do `Command::new("ngspice")` and rely on system path?

crates/pcb/src/sim.rs
```diff
@@ -1,26 +1,29 @@
...
 
-    /// Setup file (e.g., voltage sources)
+    /// Setup file (e.g., voltage sources). Only valid when simulating a single file.
     #[arg(long, value_hint = clap::ValueHint::FilePath)]
     pub setup: Option<PathBuf>,
 
-    /// Output file (use "-" for stdout)
-    #[arg(short, long, default_value = "sim.cir", value_hint = clap::ValueHint::FilePath)]
-    pub output: PathBuf,
+    /// Write .cir to a specific file (ngspice still runs on it). Only valid when simulating a single file.
```
akhilles: Either help text is wrong here or the behavior needs to be updated. If `output` is set, ngspice doesn't run as far as I can tell.

crates/pcb/src/sim.rs
```diff
@@ -30,24 +33,22 @@ pub struct SimArgs {
...
-    } else {
-        Box::new(File::create(path)?)
-    })
-}
-
-pub fn execute(args: SimArgs) -> Result<()> {
-    crate::file_walker::require_zen_file(&args.file)?;
+    /// Print the .cir netlist to stdout (skip running ngspice)
+    #[arg(long = "netlist")]
+    pub netlist: bool,
```
akhilles, line 39: If --netlist and --output <FILE> are passed together, --output is silently ignored. If this shouldn't happen, suggest adding a `conflicts_with`. But currently `--netlist` does the same thing as `-o -`, so maybe we dont need both.

crates/pcb/src/sim.rs
```diff
@@ -57,15 +58,133 @@ pub fn execute(args: SimArgs) -> Result<()> {
...
+    let zen_files = file_walker::collect_workspace_zen_files(
+        args.path.as_deref(),
+        &resolution_result.workspace_info,
+    )?;
+
+    let mut simulated = 0u32;
+    let mut has_errors = false;
+
+    for zen_path in &zen_files {
+        match simulate_one(zen_path, resolution_result.clone(), &args) {
```
akhilles, line 171: This can print all the netlists with no delimiter. Is that expected? Or should multi-file only be supported when running with ngspice. Maybe should bail early on `--netlist` like we do for `--output`

crates/pcb/src/sim.rs
hexdae: The initial idea was to make this more cohesive with `build` but keep backward compatibility: `pcb sim -o -` should now be `pcb simulate --netlist` 

Adding the conflict is a good idea.

crates/pcb/src/sim.rs
hexdae: Yeah I think if the goal is to produce a `.cir` file we should not run ngspice. I will update the text

crates/pcb-sim/src/ngspice.rs
hexdae: Pretty sure this would not work on windows, where system paths are funky depending on the execution context (wsl vs powershell and what not)

Maybe since this is a new behavior we can be conservative, but I would rather do this for now and then figure out how to execute spice sim without relying on system ngspice

crates/pcb-zen/src/lsp/mod.rs
hexdae: Addressed

crates/pcb/src/sim.rs
hexdae: Changed to disallow netlist or output unless an individual file is selected

hexdae: This is I guess different from `pcb build` but maybe we should mirror this behavior there as it is a bit more intuitive to me that you would not want multiple net lists json files concatenated with diag text

PR #592, opened 2026-02-28
[SIM] fstring net support

hexdae:
# Description

Simpler `Display` trait for net and better f-string support so we can do something like this 

```python
Simulation(
    name="test_SallenKey",
    setup=f"""
* Ground node

* Positive and negative rails
V1 {VP} {GND} DC 12V 
V2 {VN} {GND} DC -12V

* Input stimulus for AC analysis. 0V - 1.8V (centered at 0.9V, 0.9V peak AC)
V3 {INPUT} {GND} DC 0.9 AC 0.9

.control
  * AC analysis from 1Hz to 10kHz with 10 points per decade
  ac dec 200 1 10g
  
  * Save output as SVG
  set hcopydevtype = svg
  
  * AC magnitude response plot
  hardcopy output/sallenkey_magnitude.svg vdb({PROBE}) title "Sallen Key AC Response" xlabel Frequency ylabel "Magnitude (dB)"
  
  * AC phase response plot
  hardcopy output/sallenkey_phase.svg vp({PROBE}) title "Sallen Key Phase Response" xlabel Frequency ylabel "Phase (degrees)"
.endc
""",
)
```

---

> [!NOTE]
> **Medium Risk**
> Bumps the `starlark-rust` git dependency and changes how `Net` values format to strings, which can affect all Starlark formatting/f-string output and related diagnostics across the toolchain.
> 
> **Overview**
> Enables cleaner SPICE `setup` authoring by making `Net` values format as just their net name in Starlark string contexts (including f-strings), rather than a debug-style `Net(name=..., id=...)` representation.
> 
> Updates the `starlark-rust` pinned git revision and refreshes snapshots/changelog to match the new net formatting (including interface error messages like `SingleNet(signal=SIG_signal)`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8f20d68d035b2b25850ca3071eb6b29bd40c2376. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,11 +11,11 @@ authors = ["Diode Computers, Inc. <founders@diode.computer>"]
 
 [workspace.dependencies]
 glam = "0.31.0"
-starlark = { git = "https://github.com/diodeinc/starlark-rust", rev = "bd12784d" }
-starlark_map = { git = "https://github.com/diodeinc/starlark-rust", rev = "bd12784d" }
-starlark_derive = { git = "https://github.com/diodeinc/starlark-rust", rev = "bd12784d" }
-starlark_syntax = { git = "https://github.com/diodeinc/starlark-rust", rev = "bd12784d" }
-allocative = { git = "https://github.com/diodeinc/starlark-rust", rev = "bd12784d" }
+starlark = { git = "https://github.com/diodeinc/starlark-rust", rev = "41c8e1b317996a7b71f482886e66f89aa006272e" }
+starlark_map = { git = "https://github.com/diodeinc/starlark-rust", rev = "41c8e1b317996a7b71f482886e66f89aa006272e" }
+starlark_derive = { git = "https://github.com/diodeinc/starlark-rust", rev = "41c8e1b317996a7b71f482886e66f89aa006272e" }
+starlark_syntax = { git = "https://github.com/diodeinc/starlark-rust", rev = "41c8e1b317996a7b71f482886e66f89aa006272e" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -11,11 +11,11 @@ authors = ["Diode Computers, Inc. <founders@diode.computer>"]
 
 [workspace.dependencies]
 glam = "0.31.0"
-starlark = { git = "https://github.com/diodeinc/starlark-rust", rev = "e80bd650" }
```
LK, line 14: can you use 41c8e1b317996a7b71f482886e66f89aa006272e instead

crates/pcb-zen-core/src/lang/net.rs
```diff
@@ -163,11 +163,7 @@ where
 
 impl<'v, V: ValueLike<'v>> std::fmt::Display for NetValueGen<V> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(
```
LK, line 166: cc @akhilles any issue you can see with this?

this won't work for cases where nets get renamed (I think net name collision and nets within submodules are the only cases where this happens today) but should be good enough for this use case.

The other option is to serialize these as a net ID with a special marker we parse out in the sim generation

Cargo.toml
hexdae: done

crates/pcb-zen-core/src/lang/net.rs
akhilles: I think its fine since the sim target module should be the root eval module. And we should probably hard error on net name collision at some point now that we have NC.

PR #647, opened 2026-03-17
perf: improve performance for large workspaces

LK:
---

<a href="https://app.devin.ai/review/diodeinc/pcb/pull/647" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

---

> [!NOTE]
> **Medium Risk**
> Refactors dependency resolution and evaluation caching paths used by build/resolve, so mistakes could cause incorrect package root/url mapping or stale evaluation state across files. Changes are mostly internal but touch core resolution and build execution paths.
> 
> **Overview**
> **Speeds up builds in large workspaces** by reusing dependency-resolution metadata and evaluation state instead of rebuilding it per file.
> 
> `resolve_dependencies` now takes an optional `input_path` so the auto-deps scan can be scoped to a specific file/package (and callers are updated accordingly). Dependency resolution also adds cached `package_roots` and a reverse `root → URL` map (via `OnceLock`/`Arc`) to avoid repeated recomputation and linear scans when formatting/resolving `package://` URIs.
> 
> Build execution is updated to prepare a shared `EvalContextConfig` + `EvalSession` (`PreparedBuildEval`) and evaluate multiple `.zen` files via `eval_with_config_and_session`, resetting only per-root state between runs. `PcbToml` now treats empty manifest files as `default()` configs, and tests are updated (plus a new regression test) to ensure branch pinning doesn’t clobber auto-added member deps.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 45f96d4b2278d8eeca8e09ce963c0038d6e7a30a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/pcb-zen/src/auto_deps.rs
+++ b/crates/pcb-zen/src/auto_deps.rs
@@ -130,6 +130,16 @@ fn push_unknown(summary: &mut Vec<(PathBuf, Vec<String>)>, path: &Path, items: V
     summary.push((path.to_path_buf(), items));
 }
 
+fn normalize_manifest_path(path: &Path, workspace_root: &Path) -> PathBuf {
+    path.canonicalize().unwrap_or_else(|_| {
+        if path.is_absolute() {
+            path.to_path_buf()
+        } else {
+            workspace_root.join(path)
+        }
+    })
+}
+
 fn can_materialize_dep(
     workspace_info: &WorkspaceInfo,
     index: &CacheIndex,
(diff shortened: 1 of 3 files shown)


crates/pcb-zen-core/src/config.rs
```diff
@@ -177,13 +177,22 @@ impl PcbToml {
             .read_file(path)
             .with_context(|| format!("failed to read {}", path.display()))?;
 
+        if content.trim().is_empty() {
```
akhilles, line 180: curious why this is needed. I thought blank `pcb.toml` files should parse fine in the existing code path. is this for performance reasons as well?

LK: Yes not a correctness thing just performance, although I suspect this is pretty negligible

PR #717, opened 2026-04-13
Add support for `pcb build --config KEY=VALUE`

akhilles:
> [!NOTE]
> **Medium Risk**
> Introduces a new input path into the evaluator and extends implicit type coercion (string→bool/int/float), which could change runtime behavior for existing builds if callers start passing inputs or rely on previous conversion failures.
> 
> **Overview**
> **Adds repeatable `--config KEY=VALUE` overrides to the CLI** for `pcb build` (and `pcb bom`), parsing overrides as string values and passing them into Zen evaluation as JSON inputs.
> 
> **Extends the `pcb_zen` evaluation API** (`eval`/`run`) to accept an inputs map and threads it through build flows and tests, alongside a small `EvalContext` refactor to lazily create/reuse the internal context value.
> 
> **Improves implicit type conversion** so string inputs can coerce to `bool`, `int`, and `float` (in addition to existing enum/net/physical conversions), and updates docs/changelog plus adds a CLI integration test covering config overrides.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 518669b52c284ac74f083fa88d4c6ab11ea0c015. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

<a href="https://app.devin.ai/review/diodeinc/pcb/pull/717" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/pcb/src/config_input.rs
+++ b/crates/pcb/src/config_input.rs
@@ -1,0 +1,93 @@
+use anyhow::{Result, bail};
+use serde_json::Value as JsonValue;
+use starlark::collections::SmallMap;
+
+pub const CONFIG_ARG_HELP: &str = "Override root config() parameters. Repeat as needed.\n\
+     Values parse as true/false, bare ints, bare floats, otherwise strings.";
+
+fn parse_config_value(raw: &str) -> JsonValue {
+    if raw.eq_ignore_ascii_case("true") {
+        return JsonValue::Bool(true);
+    }
+
+    if raw.eq_ignore_ascii_case("false") {
+        return JsonValue::Bool(false);
+    }
+
+    if let Ok(value) = raw.parse::<i32>() {
+        return JsonValue::Number(value.into());
+    }
+
+    if let Ok(value) = raw.parse::<f64>()
+        && let Some(number) = serde_json::Number::from_f64(value)
+    {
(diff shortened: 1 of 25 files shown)


crates/pcb/src/config_input.rs
```diff
@@ -0,0 +1,93 @@
...
+fn parse_config_value(raw: &str) -> JsonValue {
+    if raw.eq_ignore_ascii_case("true") {
+        return JsonValue::Bool(true);
+    }
+
+    if raw.eq_ignore_ascii_case("false") {
+        return JsonValue::Bool(false);
+    }
+
+    if let Ok(value) = raw.parse::<i32>() {
```
LK: Hm this is tricky because if we do something like `--config package=0402` this will get parsed as an int and will fail type checking. I think the value conversion needs to be type-aware to the signature of the module?

akhilles: Im thinking of just having `config()` try string to scalar conversion. Would really like to avoid doing multiple eval passes.

akhilles: done, added a test for 0402

LK (approved):

PR #762, opened 2026-04-25
Make `pcb info`, dirty detection much faster

akhilles:
before:
```
pcb info -f json > /dev/null
1.866 s ± 0.140 s
```

after:
```
target/release/pcb info -f json > /dev/null
223.9 ms ± 3.1 ms
```

This also makes publish faster because it relies in info + dirty detection.

---

> [!NOTE]
> **Medium Risk**
> Reworks dirty-package and tag metadata discovery using git log/diff/status and batched tag reads, which could change which packages are flagged dirty/published. Touches `pcb publish` preflight ordering and workspace discovery signature across many commands, so regressions would affect CLI workflows broadly.
> 
> **Overview**
> **Improves `pcb info` and publish performance** by refactoring workspace discovery to always enrich packages with `version`, `published_at`, and `dirty` using *git metadata* instead of per-package content hashing.
> 
> This adds new git helpers (`decorated_commits`, `changed_paths_since_in_repo`, `status_paths_in_repo`) and replaces tag annotation/timestamp parsing with a batched `git cat-file` implementation to fetch tag/commit timestamps efficiently.
> 
> As a result, `get_workspace_info` drops the `enrich_versions` flag and all call sites are updated; `pcb info` no longer separately calls `populate_dirty`, `pcb publish` now fetches remote tags/branch before loading workspace info, and the JSON snapshot updates to reflect the new dirty computation behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 121a4f407918bd9f72a65a26b78b66017959f3b3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

<a href="https://app.devin.ai/review/diodeinc/pcb/pull/762" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/pcb-zen/src/workspace.rs
+++ b/crates/pcb-zen/src/workspace.rs
@@ -80,6 +80,7 @@ fn discover_dirty_packages(
     latest_tags: &BTreeMap<String, PackageTagInfo>,
     tag_metadata: &HashMap<String, git::TagMetadata>,
     status_paths: Vec<PathBuf>,
+    workspace_subpath: Option<&Path>,
 ) -> BTreeSet<String> {
     let mut dirty = BTreeSet::new();
     for url in workspace.packages.keys() {
@@ -90,14 +91,14 @@ fn discover_dirty_packages(
 
     if let Some(base) = latest_package_tagged_ref(&workspace.root, latest_tags, tag_metadata) {
         for path in git::changed_paths_since_in_repo(&workspace.root, &base) {
-            if let Some(url) = package_url_for_path(workspace, &path) {
+            if let Some(url) = package_url_for_path(workspace, &path, workspace_subpath) {
                 dirty.insert(url);
             }
         }
     }
 
     for path in status_paths {
-        if let Some(url) = package_url_for_path(workspace, &path) {
+        if let Some(url) = package_url_for_path(workspace, &path, workspace_subpath) {
             dirty.insert(url);
         }
     }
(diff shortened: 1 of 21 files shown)


crates/pcb-zen/src/workspace.rs
```diff
@@ -106,87 +93,144 @@ impl WorkspaceInfoExt for WorkspaceInfo {
...
-    let current_manifest = std::fs::read_to_string(pkg_dir.join("pcb.toml"))
-        .ok()
-        .map(|content| compute_manifest_hash(&content));
+    if let Some(base) = latest_package_tagged_ref(&workspace.root, latest_tags, tag_metadata) {
+        for path in git::changed_paths_since_in_repo(&workspace.root, &base) {
+            if let Some(url) = package_url_for_path(workspace, &path) {
+                dirty.insert(url, DirtyReason::Modified);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
akhilles: we dont allow staggered publishing, this is a risk we have to take for much better performance.

LK (approved):

crates/pcb/tests/snapshots/info__json_format_with_published_at.snap
```diff
@@ -30,7 +30,6 @@ Exit Code: 0
       "version": "0.2.0",
       "published_at": "<PUBLISHED_AT>",
       "preferred": false,
-      "dirty": true,
```
LK, line 33: why did this change?

LK (approved):

crates/pcb/tests/snapshots/info__json_format_with_published_at.snap
akhilles: It was dirty before because the test creates plain tags without the canonical hash payload, so the old code treated them as "legacy tags" and marked the package dirty. Now `pcb info` just checks for git changes since the latest package tag, so it correctly omits dirty.

I got rid of the whole "legacy tags" concept, we dont need it anymore

PR #786, opened 2026-05-08
feat: add external dependencies to `pcb info`

LK:
> [!NOTE]
> **Medium Risk**
> Medium risk because it changes `pcb info` to run full dependency resolution and emit new JSON/human output fields, which could affect tooling that parses `pcb info` output or relies on prior behavior.
> 
> **Overview**
> `pcb info` now resolves the workspace via the shared dependency resolver and reports the full *transitive* set of **external dependencies**.
> 
> JSON output adds an `external_dependencies` map keyed by `module@version`, emitting workspace-aligned package metadata (manifest `config`, entrypoints/symbol discovery, and optional lockfile hashes) and the human-readable output adds an **External dependencies** section including the dependency source (`vendor`/`cache`/`patch`).
> 
> Tests and snapshots are updated/added to validate external dependency closure (including a sum-free/locked run that must not create `pcb.sum`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5701a240a7ce0448cd4160f39f35d03573d5853b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/pcb/tests/snapshots/info__json_with_external_dependencies.snap
+++ b/crates/pcb/tests/snapshots/info__json_with_external_dependencies.snap
@@ -18,9 +18,7 @@ Exit Code: 0
   },
   "packages": {
     "workspace": {
-      "module_path": "workspace",
       "rel_path": "",
-      "source": "workspace",
       "config": {
         "workspace": {
           "pcb-version": "0.3"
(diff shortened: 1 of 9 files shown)


crates/pcb/tests/snapshots/info__json_with_external_dependencies.snap
```diff
@@ -0,0 +1,64 @@
...
+      "source": "cache",
+      "preferred": false,
+      "entrypoints": [
+        "Leaf.zen"
+      ]
+    },
+    "github.com/vendor/components/Thing@1.0.0": {
+      "module_path": "github.com/vendor/components/Thing",
+      "version": "1.0.0",
+      "rel_path": ".pcb/cache/github.com/vendor/components/Thing/1.0.0",
```
akhilles, line 45: do we need `module_path`, `rel_path`, `source`? I think these are all strictly derivable from the package uri right? would prefer to limit sources of truth if possible

LK: removed, kept just `rel_path`

LK: (which I think we have in `pcb info` now)

akhilles (approved):

PR #832, opened 2026-05-20
Split pcb shim and pcbc toolchain releases

akhilles:
`pcb` is a small shim and version manager. `pcbc` is the real compiler/toolchain. End users normally run `pcb`; the shim selects a `pcbc` toolchain, installs it if needed, and execs it as early as possible.

## Versioning model

- `pcbc` uses the workspace/toolchain version, currently the `0.3` lane.
- `pcb` has its own slower shim version, currently `0.1.0`.
- `pcb.toml` declares the desired toolchain lane with `[workspace].pcb-version`, such as `"0.3"` or explicitly `"nightly"`.
- Semver lanes never auto-upgrade across breaking lanes. A workspace on `0.3` stays on latest compatible `0.3.x`; moving to `0.4` requires an explicit manifest edit or `pcb migrate`.

## Toolchain selection

- `pcb +0.3 ...` uses the newest installed `0.3.x`, or installs the newest available `0.3.x` if none is installed.
- `pcb +0.3.83 ...` uses exactly `0.3.83`.
- `pcb +latest ...` uses the newest stable `pcbc` release.
- `pcb +nightly ...` uses the moving nightly channel built from `main`.
- Without a CLI override, `pcb` walks up to find `pcb.toml` and uses `[workspace].pcb-version`; without a manifest it falls back to `latest`.

## Release channels

- `v0.3.84` tags publish stable `pcbc` releases via `.github/workflows/pcbc-release.yml`.
- `pcb/v0.1.1` tags publish shim releases via `.github/workflows/pcb-release.yml` and update `pcb-latest.json`.
- Nightlies are published by `.github/workflows/pcbc-nightly.yml` on a schedule from `main`. If the current `main` SHA already has a nightly, the workflow exits without rebuilding.
- Release artifacts are raw target-named binaries plus `.sha256` files in S3. Stable releases are discovered from S3 XML listings under `pcb/v*`; moving channels use small JSON pointers.

## Local cache layout

```text
<data-dir>/pcb/toolchains/
  0.3.83/<target>/pcbc
  nightly/<target>/pcbc
```

Stable installs are keyed by semver. Nightly is keyed by channel and records the date/SHA in `receipt.json`.

## `pcb` subcommands

Shim-owned commands:

- `pcb self update` updates the shim and installed managed toolchains.
- `pcb toolchain list` shows installed stable and nightly toolchains.
- `pcb toolchain show` shows the active toolchain for the current directory.
- `pcb toolchain install <request>` installs `0.3`, `0.3.83`, `latest`, or `nightly`.
- `pcb toolchain uninstall <version|nightly>` removes an installed toolchain.

All other commands are implemented by `pcbc` and reached through the shim.

---

> [!NOTE]
> **High Risk**
> High risk because it rewires the primary `pcb` entrypoint into a networked shim that installs/execs versioned toolchains and replaces the existing release/self-update workflows, which can break installs, upgrades, or command routing across platforms.
> 
> **Overview**
> **Splits the CLI into two deliverables:** `pcb` becomes a shim/version manager (now `0.1.0`) that selects a `pcbc` toolchain based on `pcb.toml`/`+<override>` and then execs it; the original full CLI functionality moves to a new `pcbc` binary/crate.
> 
> **Adds toolchain install/update mechanics in the shim:** `pcb` now supports `self update` and `toolchain {list,show,install,uninstall}`, downloads platform-specific artifacts from the Diode S3/CDN with SHA256 verification, caches release listings, and manages per-version and nightly installs with locking.
> 
> **Replaces release automation:** removes the old `cargo-dist`/main build/S3 upload workflows and adds dedicated workflows for `pcbc` releases (`v*` tags), `pcb` shim releases (`pcb/v*` tags + `pcb-latest.json`), scheduled `pcbc` nightlies (with `nightly/latest.json` pointer), and separate version-bump dispatch workflows; updates docs/tests/README accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2db29c10550637cfb113beffd5d21c0b9505095b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crates/pcbc/Cargo.toml
```diff
@@ -0,0 +1,87 @@
...
+jiff = { workspace = true }
+crossterm = { workspace = true }
+ctrlc = { workspace = true }
+reqwest = { workspace = true }
+termimad = { workspace = true }
+syntect = { workspace = true }
+gltfpack-sys = { workspace = true }
+thiserror = { workspace = true }
+minijinja = { workspace = true }
+canon-json = "0.2"
```
LK, line 74: the one lone non-`workspace` dependency 😢

LK (approved):
lgtm overall, couple small nits:

1. `pcb --version` gives me the shim version, `pcb +0.83.0 --version` gives me the pcb version. would expect `pcb --version` to give me the version that `pcb toolchain show` gives me
2. `pcb self update` 404s because we have no releases published with the new release pipeline yet?
3. we should update `pcb-action` accordingly so we preload the correct `pcb` version when we do an install
4. (as discussed offline): the cargo-dist generated github releases were nice because they included install commands inline, lets set up install scripts (windows + unix) in the repo and make the generated releases include the same install scripts. we should also update `docs.pcb.new` to point to these

PR #881, opened 2026-06-12
Improve remote layout sync recovery

LK:
Perform JWT refresh during layout sync session, be more lenient with heartbeat retries, automatically soft-quit KiCad on sync disconnect, and proactively suggest restoring previous session when reconnecting. This should make the layout syncing more stable in the short term

---

> [!NOTE]
> **Medium Risk**
> Changes remote file upload format, lock heartbeat semantics, and KiCad shutdown behavior during sync failures; incorrect recovery or terminate logic could strand locks or overwrite remote layouts.
> 
> **Overview**
> Improves **remote sandbox layout sync** so interrupted KiCad sessions are less likely to lose work and long sessions stay authenticated.
> 
> **Sandbox API client** now uploads layout files as raw `application/octet-stream` instead of base64 JSON, resolves the bearer token through `WorkspaceContext` on each request (so tokens can refresh mid-session), tightens lock heartbeats (5s interval, up to three transient failures before the lock is treated as inactive), and treats a missing or mismatched lock lease as “lost” rather than failing immediately.
> 
> **`pcb open` / remote layout** adds per-session local caches with a `.pcb-sync-session.json` manifest, optional interactive restore of the latest recoverable local board after a bad exit, retries uploads three times, and on recoverable stops (lock reclaimed or sync errors) keeps the local recovery copy, marks the session recoverable, and attempts to quit KiCad via `PcbnewSession::terminate()` before releasing the lock. Successful exits still do a final upload; messaging points users at the on-disk recovery file.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd950ae594e20f95b1c3b281735bd8163ae9b2a2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/pcbc/src/remote_sandbox.rs
+++ b/crates/pcbc/src/remote_sandbox.rs
@@ -1,10 +1,12 @@
 use anyhow::{Context, Result, bail};
+use chrono::{DateTime, Utc};
+use inquire::Confirm;
 use notify::{EventKind, RecommendedWatcher, RecursiveMode, Watcher};
 use pcb_diode_api::{
     ExecSyncRequest, SandboxClient, SandboxFileUri, SandboxLockGuard, SandboxLockOptions,
 };
 use rayon::prelude::*;
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use std::collections::BTreeSet;
 use std::fs;
 use std::path::{Path, PathBuf};
@@ -14,7 +16,7 @@ use std::sync::{
     mpsc::{self, Receiver, RecvTimeoutError},
 };
 use std::thread;
-use std::time::Duration;
+use std::time::{Duration, SystemTime};
 use uuid::Uuid;
 
 use crate::layout::{LayoutArgs, LayoutOutputFormat};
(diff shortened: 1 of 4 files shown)


crates/pcbc/src/remote_sandbox.rs
```diff
@@ -412,16 +630,287 @@ fn sync_layout_up(
...
+            .as_ref()
+            .is_none_or(|current| session.manifest.updated_at > current.manifest.updated_at)
+        {
+            latest = Some(session);
+        }
+    }
+    Ok(latest)
+}
+
+fn prompt_restore_recovery(
```
akhilles: this isn't fallible afaik, don't need Result wrapping

crates/pcbc/src/remote_sandbox.rs
```diff
@@ -95,6 +102,8 @@ struct LocalLayout {
     remote_layout_dir: String,
     local_layout_dir: PathBuf,
     pcb_file: PathBuf,
+    sync_session: Option<SyncSession>,
+    restored_from: Option<PathBuf>,
```
akhilles: this path should always match `pcb_file` right? can this just be a bool? to avoid possibility for these not agreeing.

akhilles (approved):

PR #903, opened 2026-06-24
feat: add `pcb rectify`

lucia3e8:
`pcb rectify check file.kicad_mod` tells you whether the included step file is aligned with the footprint
`pcb rectify fix file.kicad_md` aligns the model to the footprint if it isn't

both of these use clever heuristics like "how many pins to I think this 3d model has and where are their pads". entire tool is courtesy of @hexdae 

it can correctly align ~70% of the registry right now

small shim change: `pcb-rectify` ships as a sidecar binary, this means the s3 upload state is no longer atomic (one of the binaries will upload first). we added sentinels in [index/](https://us-west-2.console.aws.amazon.com/s3/buckets/diode-pcb-releases?region=us-west-2&prefix=pcb/index/&showversions=false) that get created after all the binaries finish uploading. the shim checks if the sentinel exists first, that way we never show partial state

---

> [!NOTE]
> **Medium Risk**
> Touches toolchain download/version discovery and multi-artifact publishing; incorrect marker or sidecar logic could block installs or serve incomplete releases, but completion markers explicitly reduce that race.
> 
> **Overview**
> Adds **`pcb rectify check`** and **`pcb rectify fix`**, implemented by a new **`rectify`** crate and the **`pcb-rectify`** sidecar. The tool tessellates STEP geometry, scores axis-aligned poses against footprint pads/holes (SMD contact vs THT pin-island paths), and can rewrite `(rotate ...)` / `(offset ...)` in `.kicad_mod` files. **`pcbc`** now resolves external commands on PATH and beside the toolchain binary so bundled sidecars work without being on PATH.
> 
> **Release & shim:** Nightly/release CI builds and publishes **`pcb-rectify`** next to **`pcbc`**. The shim stages optional sidecars best-effort (marker file avoids re-download loops) and only lists versions with a **`pcb/index/v<semver>`** completion marker written after full S3 upload, replacing the old “stdlib first” partial-release exposure.
> 
> Hidden **`pcb-rectify`** subcommands (`solve`, `patch`, `bench`, `audit`) support benchmarking and tuning; workspace deps pull in foxtrot STEP tessellation, FFT/raster numerics, etc.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f05873fd6d08461548f1889c767f8e58a163849. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crates/rectify/src/main.rs
```diff
@@ -0,0 +1,354 @@
...
+        fail_on_flagged: true,
+        mode: if args.strict {
+            bench::BenchMode::Strict
+        } else {
+            bench::BenchMode::Loose
+        },
+        randomize_initial_transform: !args.use_stored_initial_transform,
+        initial_transform_seed: args.initial_transform_seed,
+    })
+}
```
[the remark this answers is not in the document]
lucia3e8: Verified non-issue: the initial transform is a no-op for the solver's result. `rectify` re-infers pose by enumerating all 24 axis-aligned rotations and recomputing the translation from pad/hole geometry — it never uses the stored/initial transform as a prior. Confirmed empirically by running `check` with and without `--use-stored-initial-transform` on several footprints (SMA_73251-1153, FTSH-105-01-L-DV-K-A-P-TR, USB4155-03-C): identical predicted rotate + offset in every case. So randomization only matters for `bench` (to prevent the benchmark from trivially preserving the file's transform) and cannot produce false failures in `check`/`fix`. Resolving.

crates/pcb/src/main.rs
```diff
@@ -779,6 +809,25 @@ fn stage_stdlib_archive(base_url: &str, staging_dir: &Path) -> Result<()> {
...
+            };
+            verify_checksum(&url, &bytes)?;
+            let dst = staging_dir.join(executable_name(binary));
+            fs::write(&dst, bytes)?;
+            copy_executable_permissions(&dst, &dst)?;
+            break;
+        }
+    }
+    Ok(())
+}
```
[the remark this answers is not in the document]
lucia3e8: Fixed in 7248b006376c5600f40409582aaf5eaec3569a64. `stage_optional_sidecars` is now best-effort and writes a `.sidecars-checked` marker once staging completes (sidecar fetched, or confirmed unavailable via all-404). `optional_sidecars_present` short-circuits on that marker, so older toolchains no longer re-acquire the install lock or re-hit the network on every command. A hard/transient failure leaves no marker so the next invocation retries.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to Semantic Versioning (https://semver.org/spec/v2.0.0.
 - Removed legacy v1/`pcb.sum` resolution, disabled `pcb update`, and dropped obsolete `--locked` read-command flags.
 - `pcb migrate` removes deprecated `[workspace].members`; other commands reject it.
 - `pcb sync` no longer writes stdlib-only KiCad dependency entries.
+- Added `pcb rectify check` and `pcb rectify fix`, backed by a bundled `rectify` sidecar binary, to check and patch KiCad footprint 3D model `(rotate ...)`/`(offset ...)` transforms by matching tessellated STEP geometry against the footprint's pads and holes.
```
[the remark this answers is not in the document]
lucia3e8: Fixed in 7248b006376c5600f40409582aaf5eaec3569a64: moved the `pcb rectify` entry from the released [0.4.0] section to [Unreleased] > ### Changed.

.github/workflows/pcbc-release.yml
akhilles, line 1: Small race condition with publishing to s3: its possible someone attempts an update/install mid-publish. If rectify hasn't been uploaded yet, `pcb` writes the marker file saying “don’t check again.” This is more of a bug in pcb shim I think. Any object under `pcb/vX.Y.Z/` makes `vX.Y.Z` appear as a release prefix. Might be a good ideal to upload a `.complete` marker last or some manifest. It should be very rare.

skills/rectify/SKILL.md
akhilles, line 1: This is very confusing but these are actually automatically synced from `https://github.com/diodeinc/diode/tree/main/projects/ai/skills`. So should be added there, CI will sync it to this repo.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to Semantic Versioning (https://semver.org/spec/v2.0.0.
 ### Changed
 
 - `NotConnected` is now an open-net constructor, not a net type.
+- Added `pcb rectify check` and `pcb rectify fix`, backed by a bundled `pcb-rectify` sidecar binary, to check and patch KiCad footprint 3D model `(rotate ...)`/`(offset ...)` transforms by matching tessellated STEP geometry against the footprint's pads and holes.
```
akhilles: Add to ### Added instead?

akhilles (approved):

.github/workflows/pcbc-release.yml
lucia3e8: gonna replace with upload to another folder then mv to final location since mv is atomic

lucia3e8: i lied s3 has no atomic `mv` we're doing the completion sentinel like you said

skills/rectify/SKILL.md
lucia3e8: makes sense, i took them out

PR #917, opened 2026-07-03
fix: document stdlib files from any workspace

LK:
## Summary
- Treat materialized stdlib paths as stdlib targets during doc resolution.
- Preserve frozen root lookup for stdlib files so `pcb doc @stdlib/...` works from any workspace.
- Add a regression test covering stdlib target resolution and frozen root selection.

## Testing
- Not run (not requested)

---

> [!NOTE]
> **Low Risk**
> Narrow change to package/doc resolution for stdlib paths with a regression test; no auth, sync, or build pipeline changes.
> 
> **Overview**
> Fixes **`pcb doc @stdlib/...`** when run from workspaces that also contain packages under `.pcb/stdlib`, by treating materialized stdlib as the **`@stdlib`** root instead of a normal workspace member.
> 
> **`target_package_urls_for_path`** now maps paths under the workspace stdlib directory to **`STDLIB_MODULE_PATH`** only. **`FrozenResolutionBuilder::build`** short-circuits for that target and builds a frozen map with a single **`FrozenPackageIdentity::Stdlib`** entry. **`frozen_root_for_file`** includes stdlib-owned files when picking the frozen root, so doc/eval can set **`active_root_package`** correctly.
> 
> Adds a regression test for stdlib target resolution and frozen root selection; changelog entry under **[Unreleased]**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1eff26347523f1820fbcf606e78008f6bad26a0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to Semantic Versioning (https://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- `pcb doc @stdlib/...` now documents stdlib files from any workspace without treating the materialized stdlib as an ordinary workspace package.
+
 ## [0.4.4] - 2026-07-03
 
 ### Added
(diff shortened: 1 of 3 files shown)


akhilles (commented):
I think https://github.com/diodeinc/pcb/pull/928 might be a better fix because stdlib files resolve to the stdlib root only when one exists, instead of matching every map.

PR #940, opened 2026-07-22
ENG-445 Fix strict BOM API request

akhilles:
## Summary

- Send strict BOM selection through the API's `strict=true` query parameter.
- Decode the complete strict BOM match-status contract.
- Document the fix under `Unreleased`.

## Why

The PCB client sent `strict` in the request body, but the Diode API reads it from the query string. As a result, `[workspace.bom] strict = true` still used the default matcher. Strict responses can also contain `MATCH_COMPATIBLE` and `MATCH_NEEDS_RETRY`, which the client did not previously decode.

## Validation

- `cargo fmt --check`
- `cargo test -p pcb-diode-api` (117 tests)
- `pnpm -C projects/api test test/routes/boms.test.ts` in `../diode` (2 tests)

---

> [!NOTE]
> **Low Risk**
> Localized BOM API client changes with added tests; behavior change is that strict mode now takes effect as users intended.
> 
> **Overview**
> Fixes **`[workspace.bom] strict = true`** so it actually enables the API’s strict BOM matcher: **`strict` moves from the JSON body to `?strict=true` on `/api/boms/match`**, which is what the Diode API reads.
> 
> The client also **aligns with the strict match contract** by decoding **`MATCH_COMPATIBLE`** and **`MATCH_NEEDS_RETRY`**, adding **`UK`** geography (filtered out of US/global availability and offer lists like other non-target regions), and **expanding tests** for URL construction, status/geography JSON, and UK offer handling. The **Unreleased changelog** documents the strict-setting fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 94ce656fbd81cfcfd35ac2ff4c9bd9554d8feda9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to Semantic Versioning (https://semver.org/spec/v2.0.0.
 
 - `pcb scan` supports page ranges with `--pages START-END`.
 
+### Fixed
+
+- `[workspace.bom] strict = true` now selects the API's strict BOM matcher instead of sending an ignored request-body field.
+
 ## [0.4.8] - 2026-07-22
 
 ### Changed
--- a/crates/pcb-diode-api/src/bom.rs
+++ b/crates/pcb-diode-api/src/bom.rs
@@ -105,8 +105,12 @@ struct DesignBomEntry {
 enum BomMatchStatus {
     #[serde(rename = "MATCH_EXACT")]
     Exact,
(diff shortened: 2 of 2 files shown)


LK (approved):
lgtm, we should toggle this on by default once we are confident with the strict BOM flow

PR #956, opened 2026-07-27
Add standalone KiCad schematic import

craigxchen:
Adds `pcb import` support for standalone `.kicad_sch` files, alongside the existing `.kicad_pro` project form, and hardens the import path.

Both input forms share one pipeline; there is no separate project-import path.

## What import does

Converts a KiCad design into a buildable Zener board repository:

```
pcb import <design.kicad_sch | project.kicad_pro> [output-directory]
```

- **Connectivity is the contract.** Every KiCad physical pin maps to its own Zener signal via `pin_defs`, so duplicate displayed pin names, separate no-connect terminals, and pinless mechanical footprints stay distinct. Import builds the result and verifies every source `(refdes, pin)` net partition before it finishes; a mismatch fails rather than shipping a wrong board.
- **Only ever adds files.** A generated path that differs from an existing one keeps the existing content and is reported, so hand-authored or agent-fixed sources survive a re-import. `--force` replaces what that run generates, and never `pcb.toml`.
- **Offline by construction.** One place resolves a workspace and it always resolves offline; MPN lookup reads only on-disk indexes; a registry substitution candidate must already be cached. Nothing downloads.
- **Degrades instead of failing.** Unresolved footprints and incomplete sourcing are retained as provenance and surface as `bom.imported_incomplete` at build time. The board still builds; it just isn't layout-ready until footprints are supplied.

Output is minimal — the converted sources plus `pcb.toml` and `.gitignore` when absent. No `git init`, no `README.md`, no vendored stdlib.

## Verification

Swept the CLI over **91 real KiCad project roots** extracted from a 12 GB SparkFun repo archive (all 129 `.kicad_sch` in the archive, verified against a full listing):

| | import ok | build ok | build failed |
|---|---|---|---|
| **result** | **69/91 (76%)** | **69** | **0** |

2,831 components and 3,135 physical-pin partitions validated. Every design that imported also built.

Built the pre-change commit as a baseline and ran the identical corpus through it:

- **69/91 both** — per-project join over all 91: **zero regressions, zero exit-code changes**.
- Design content (`.zen`, `.kicad_sym`, `.kicad_mod`, `pcb.toml`, `.gitignore`) **byte-identical on 69/69**. The only output differences are intended: diagnostics relocated to `.pcb/import/`, `README.md` no longer written, source archive now opt-in.

Also driven at the CLI: re-import idempotence (silent when nothing changed), hand-edit preservation, `--force` overwrite, the `.kicad_pro` form with `layout/`, bad-input handling, and byte-identical output under an OS sandbox denying all network access.

## Known limitations

The 22 sweep failures are pre-existing, confirmed by baseline parity and `git log -S`. Grouped:

- **14 — the netlist references a pin the schematic's embedded symbol doesn't define.** KiCad exports the netlist against the *external* symbol library while pin extraction reads the schematic's embedded `lib_symbols` copy; when those disagree (e.g. an exposed pad absent from the cached copy) import refuses rather than dropping a connection. Fix direction: fall back to the external symbol library for pin definitions, as extraction already does for geometry.
- **6 — a PCB footprint with no schematic symbol instance** (silkscreen logos, stray passives). Predates this branch.
- **2 — a built component that cannot be mapped back to a source reference designator.**

Coverage gap worth knowing: most SparkFun designs ship no project `fp-lib-table` and reference a globally-installed library, so footprints did not resolve on the test machine. That leaves footprint resolution, passive promotion (gated on a 2-pad footprint), and registry substitution unexercised *by that corpus* — they are covered instead by two local designs that carry their own libraries.

Registry substitution stays rare in practice for a separate reason: it requires the candidate module to declare `Part(...)`, and only 37 of 544 cached registry component entrypoints do.

## Notes for review

- `crates/pcbc/src/import/README.md` documents the pipeline, the write policy, the offline guarantee, and the limitations above.
- `.claude/skills/verify/SKILL.md` records the build-and-drive recipe, including two traps that cost a re-run each: the binary must stay inside the repo tree to locate the stdlib, and macOS has no `timeout`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **High Risk**
> Large changes to KiCad import, connectivity (`pin_defs`), footprint resolution, and post-import electrical validation—errors could silently or loudly break real designs.
> 
> **Overview**
> **`pcb import`** now takes either a **`.kicad_sch`** or **`.kicad_pro`**. Standalone runs skip layout/archive output and resolve footprints from project/global/bundled/cached libraries (with warnings when geometry is missing); project imports keep the existing layout and archive behavior.
> 
> **Component generation** gains an explicit per-pin plan: **`pcb-component-gen`** can emit **`pin_defs`**, optional **`footprint`**, and JSON-literal Starlark strings so each physical pad stays a distinct Zener signal (duplicate display names, separate NC pins, pinless mechanical parts).
> 
> The import pipeline **stages KiCad sources in a temp dir** for validation/netlist work, **atomically replaces** generated `.kicad_sym`/`.kicad_mod` (no symlink follow), and **`--force`** cleanup is source-kind aware (standalone preserves user `layout/`, drops stale archives). **`generated_validate`** offline-builds the board and compares physical-pin sets and net partitions to the KiCad source.
> 
> Smaller fixes: disabled **`fp-lib-table`** rows ignored like KiCad, ambiguous netlist identities fail with clearer errors, footprint separator preserved for generated names, and passive promotion only when symbol pins are actually `1`/`2`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aeef21d6435600b5328fa21204c95fcd47cf90f1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crates/pcbc/src/import/flow.rs
```diff
@@ -206,13 +306,19 @@ impl Analyzed {
...
+            mut ir,
         } = hierarchized;
 
-        let semantic = semantic::analyze(&ir);
+        registry_lookup::inherit_embedded_symbol_properties(
+            &mut ir.components,
+            &ir.schematic_lib_symbols,
+        );
+        let mut semantic = semantic::analyze(&ir);
+        semantic.registry_mpn_lookup =
```
akhilles: Can we drop all registry lookup and reuse here? Import should always generate local components. Otherwise the result depends on whatever is cached on that machine.

crates/pcbc/src/import/output.rs
```diff
@@ -0,0 +1,703 @@
...
+    builder.prefix(prefix);
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        builder.permissions(fs::Permissions::from_mode(0o700));
+    }
+    builder.tempdir()
+}
+
+/// Writes the import output directly into its destination directory.
```
akhilles: Can we keep the existing refuse or `--force` behavior? The new keep, package, and rollback state is a lot of machinery for a separate re-import feature.

akhilles (commented):
Can we trim this down to standalone `.kicad_sch` import, footprint lookup, pin mapping, and validation? I'd drop registry lookup and reuse, additive output handling, the archive flag, scaffolding changes, and the new BOM warning from this PR.

crates/pcb-zen-core/src/convert.rs
```diff
@@ -441,6 +441,18 @@ impl ModuleConverter {
                 .reference_designator
                 .as_deref()
                 .unwrap_or(instance.type_ref.module_name.as_ref());
+            if instance.boolean_attr(&[crate::attrs::IMPORTED_BOM_SOURCE_INCOMPLETE]) == Some(true)
```
akhilles: Can we drop this? An importer flag shouldn't change BOM errors globally. Import validation can suppress the existing error locally.

crates/pcbc/src/import/types.rs
```diff
@@ -94,36 +94,49 @@ impl From<String> for KiCadLibId {
...
     #[arg(value_name = "OUTPUT_DIR", value_hint = clap::ValueHint::AnyPath)]
-    pub output_dir: PathBuf,
+    pub output_dir: Option<PathBuf>,
 
-    /// Overwrite generated files if the board repository already exists
+    /// Overwrite existing files that this import regenerates; without it, colliding files are kept
     #[arg(long = "force")]
     pub force: bool,
+
+    /// Also write a portable archive of the KiCad sources next to the imported board
```
akhilles: Can we drop `--archive-sources` and keep the existing project import behavior? This is unrelated to `.kicad_sch` support.

crates/pcbc/src/new.rs
```diff
@@ -313,6 +362,39 @@ fn execute_new_board(board: &str, repo: &str) -> Result<()> {
     Ok(())
 }
 
+/// Write the minimum a board directory needs to be buildable: the workspace manifest, plus the
+/// `.gitignore` that keeps `.pcb/` — where `pcb build` and `pcb import` put their untracked output —
+/// out of `git status`. Neither file is written when it already exists.
+///
+/// Used by `pcb import`, which must not scaffold a repository around the design it converts: no
```
akhilles: Can we keep the existing import scaffolding? Dropping git init, the README, and stdlib setup is a separate behavior change.

crates/pcbc/src/import/materialize.rs
```diff
@@ -2,32 +2,69 @@ use super::*;
 use anyhow::{Context, Result};
 use pcb_zen_core::Diagnostics;
 use serde::Serialize;
-use std::fs;
 use std::path::{Path, PathBuf};
 
+/// Name of the validation diagnostics file `materialize` writes on every import, inside
```
akhilles: Can we leave the report paths and failure behavior alone? Moving them under `.pcb/import` and making writes nonfatal is unrelated.

crates/pcbc/src/import/extract.rs
```diff
@@ -568,6 +588,361 @@ fn extract_kicad_layout_data(
...
+            sheetname: None,
+            sheetfile: None,
+            attrs: Vec::new(),
+            properties: BTreeMap::new(),
+            pads: resolved.0.clone(),
+            footprint_geometry: resolved.1.clone(),
+        });
+    }
+
+    if !unresolved.is_empty() {
```
akhilles, line 722: Can this just print a short warning with a capped list of missing footprint IDs? The ranking and library diagnosis feel like overkill.

crates/pcbc/src/import/generate.rs
```diff
@@ -2008,86 +2330,516 @@ fn patch_symbol_footprint_property(library_text: &str, footprint_stem: &str) ->
...
+        .flatten();
+
+    ImportedSourcing {
+        mpn,
+        manufacturer_from_registry: manufacturer.is_some(),
+        manufacturer,
+    }
 }
 
 fn render_component_zen(
```
akhilles, line 2593: Can we reuse the `pcb-component-gen` renderer here and just pass in the import pin plan? We shouldn't maintain two `Component(...)` renderers.

crates/pcbc/src/new.rs
```diff
@@ -313,6 +362,39 @@ fn execute_new_board(board: &str, repo: &str) -> Result<()> {
...
+/// `git init`, no `README.md`, and no template board `.zen`, because generation writes the real one.
+pub(crate) fn write_board_scaffolding(dir: &Path, board: &str) -> Result<()> {
+    let pcb_toml = dir.join("pcb.toml");
+    if !pcb_toml.exists() {
```
akhilles: `exists()` misses dangling symlinks, and `write` will follow one here. Please use the same symlink-safe write path as the rest of import.

crates/pcbc/src/import/extract.rs
```diff
@@ -13,13 +13,20 @@ pub(super) fn extract_ir(
...
+            &paths.kicad_project_root,
+            &validation.summary.selected,
+        )?
+    } else {
+        BTreeMap::new()
+    };
 
+    // KiCad netlist export can update project-local preference files. Run it against the same
+    // contained source copy used for validation, then parse the original source files read-only.
+    let netlist_sources = portable::stage_project_files(&selection.portable)?;
```
akhilles: We stage once for ERC, again here for netlist export, then read the schematic and PCB from the originals. That can mix snapshots if the source changes during import. Can we stage once and use that tree for the whole pipeline?

crates/pcbc/src/import/types.rs
```diff
@@ -679,12 +712,11 @@ pub(super) struct ImportValidation {
 pub(super) struct GeneratedArtifacts {
     pub(super) board_dir: PathBuf,
     pub(super) board_zen: PathBuf,
-    pub(super) validation_diagnostics_json: PathBuf,
-    pub(super) import_extraction_json: PathBuf,
     pub(super) layout_dir: PathBuf,
-    pub(super) layout_kicad_pro: PathBuf,
-    pub(super) layout_kicad_pcb: PathBuf,
-    pub(super) portable_kicad_project_zip: PathBuf,
+    pub(super) layout_kicad_pro: Option<PathBuf>,
```
akhilles, line 720: This drops `validation_diagnostics_json` and `import_extraction_json` from the report even though both files still exist. Can we keep those fields and only make the layout and archive fields optional?

crates/pcbc/src/import/report.rs
```diff
@@ -50,11 +50,32 @@ pub(super) fn build_import_report(
...
 pub(super) fn write_import_extraction_report(
-    board_dir: &Path,
+    out_path: &Path,
     payload: &super::ImportReport,
-) -> Result<PathBuf> {
-    let out_path = board_dir.join(".kicad.import.extraction.json");
-    fs::write(&out_path, serde_json::to_string_pretty(payload)?)
+) -> Result<()> {
+    let mut value = serde_json::to_value(payload)?;
+    sort_json_objects(&mut value);
```
akhilles: `serde_json::Map` is already sorted since `preserve_order` isn't enabled. Isn't this recursive sort a no-op? Can we drop it and serialize the report directly?

crates/pcbc/src/import/generate.rs
```diff
@@ -2008,71 +2210,396 @@ fn patch_symbol_footprint_property(library_text: &str, footprint_stem: &str) ->
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct PhysicalPinPlan {
+    bindings: Vec<ImportComponentPinBinding>,
+    io_pins: BTreeMap<String, KiCadPinNumber>,
+}
+
+#[derive(Debug, Default)]
+struct PhysicalPinTypeMetadata {
```
akhilles: This repeats the electrical type and alternate-pin scan already in `pcb-component-gen`. Can we share a small helper instead of keeping both versions?

crates/pcbc/src/import/types.rs
```diff
@@ -94,15 +94,15 @@ impl From<String> for KiCadLibId {
...
-    pub kicad_pro: PathBuf,
+    /// Path to a KiCad schematic (.kicad_sch) or project (.kicad_pro)
+    #[arg(value_name = "KICAD_INPUT", value_hint = clap::ValueHint::AnyPath)]
+    pub kicad_input: PathBuf,
 
-    /// Output directory (a board repository will be created if needed)
+    /// Output directory; defaults to ./<input-file-stem>
     #[arg(value_name = "OUTPUT_DIR", value_hint = clap::ValueHint::AnyPath)]
-    pub output_dir: PathBuf,
+    pub output_dir: Option<PathBuf>,
```
akhilles: Can we keep `OUTPUT_DIR` required here? Making it optional is a separate CLI change and adds the default-path and extra name-validation code.

akhilles (commented):
A few more from the latest pass.

craigxchen:
Something seems wrong w the windows test we keep getting kicad missing DLL error @akhilles

crates/pcbc/tests/import.rs
```diff
@@ -0,0 +1,635 @@
...
+
+    assert!(!import.status.success());
+    let stderr = String::from_utf8_lossy(&import.stderr);
+    assert!(
+        stderr.contains("<OUTPUT_DIR>"),
+        "unexpected stderr:\n{stderr}"
+    );
+}
+
+#[test]
```
akhilles, line 65: This wrapper is what's breaking Windows CI, not the importer. Can we keep this regression test non-Windows like the other import tests and drop the wrapper platform code?

crates/pcbc/src/import/paths.rs
```diff
@@ -32,17 +41,15 @@ fn ensure_board_repo_root(path: &Path) -> Result<PathBuf> {
     if path.exists() && !path.is_dir() {
         anyhow::bail!("Output directory is not a directory: {}", path.display());
     }
-    if !path.exists() {
-        fs::create_dir_all(path)
-            .with_context(|| format!("Failed to create output directory: {}", path.display()))?;
-    }
-    let workspace_root = fs::canonicalize(path).unwrap_or_else(|_| path.to_path_buf());
+    let workspace_root = if path.exists() {
```
akhilles: Do we need to change output creation here? The old `create_dir_all` plus `canonicalize` flow was simpler. This custom missing-path handling feels unrelated to standalone schematic support.

CHANGELOG.md
```diff
@@ -8,12 +8,26 @@ and this project adheres to Semantic Versioning (https://semver.org/spec/v2.0.0.
...
+- Standalone KiCad imports keep enabled project footprint libraries authoritative when a referenced footprint is missing.
+- KiCad imports preserve the footprint separator used to disambiguate generated component names.
+
 ## [0.4.14] - 2026-07-28
 
 ### Added
 
 - Added `pcb auth git configure` and `pcb auth git unconfigure` to manage host-specific global Git credential configuration for DiodeHub.
 - PCB now authenticates its own DiodeHub Git operations without requiring global Git configuration.
+- `pcb import` accepts a standalone `.kicad_sch` as well as a `.kicad_pro`.
```
akhilles: These changes are landing after 0.4.14, so they should be under `Unreleased`.

akhilles (commented):
The PR description is still from the old version. It says output is optional, writes are additive, registry substitution happens, scaffolding is minimal, reports moved, and archives are opt-in. Can we trim it to what this PR actually does now?

crates/pcbc/tests/import.rs
craigxchen: Done in 5a802b04: the whole import test file is now `#![cfg(not(target_os = "windows"))]` like the other pcbc integration tests, and the wrapper is back to a plain `#!/bin/sh` script with no platform code.

crates/pcbc/src/import/paths.rs
craigxchen: Done in 41b88f76: restored the original `create_dir_all` + `canonicalize` flow and removed the custom missing-path resolution.

CHANGELOG.md
craigxchen: Done in e82e7f5c: moved the import entries from the 0.4.14 section to `Unreleased`; 0.4.14 now matches what actually shipped.

crates/pcbc/src/import/extract.rs
craigxchen: Done in 5b43e2ff (refactor(import): reuse staged source snapshot).

crates/pcbc/src/import/types.rs
craigxchen: Done in b7f0da89 (fix(import): restore generated report artifacts).

crates/pcbc/src/import/generate.rs
craigxchen: Done in f95c5887 (refactor: share no-connect pin classification).

crates/pcbc/src/import/types.rs
craigxchen: Done in 3e4a0951 (fix(import): require output directory).

crates/pcbc/src/import/report.rs
craigxchen: Done in b7f0da89: dropped the recursive sort and the report serializes directly (serde_json::Map is already ordered without preserve_order).

akhilles (approved):

PR #1011, opened 2026-08-04
Adopt ruff 0.16's default lint rule set for the layout scripts

akhilles:
Follow-up to the dependency sweep (#1010), which bumped ruff 0.15.20 -> 0.16.1 and pinned the old default rule selection to keep that bump behaviour-preserving. This removes the pin and adopts ruff 0.16's full default selection.

`ruff check` is now clean — all 420 findings resolved.

## What changed

Mostly mechanical: PEP 585/604 annotations, import sorting, redundant `.keys()`, collapsed nested `if`s, `ClassVar` on a mutable class default, explicit `Optional` on implicit-optional parameters.

Five error-handling sites got `# noqa` with a stated reason instead of a rewrite — changing error handling in the KiCad sync path is not a lint change. **Two are marked FIXME** because they genuinely swallow failures with no logging and deserve a real fix:

- `lens/kicad_adapter.py` — fragment discovery failures silently `return`
- `update_layout_file.py` — connectivity-rebuild failures silently `pass`

## The constraint worth knowing about

**These scripts run inside KiCad's bundled interpreter, which is Python 3.9.13 as of KiCad 10** — not the `requires-python = ">=3.10"` in `pyproject.toml`, which only describes the uv dev/test environment.

ruff derives its target from `requires-python`, so the first attempt rewrote `Optional[X]` to PEP 604 `X | None` in 40 places across modules KiCad imports. That raises at import time on 3.9:

```
layout_path: str | None = None
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

I caught this before committing, but it is worth stating plainly that **neither `pytest` nor `ty check` catches it** — both run on Python 3.12.

The fix is two settings, both commented in `pyproject.toml` so they do not get "helpfully" deleted later:

- `target-version = "py39"` — pins ruff to the real runtime floor
- `required-imports = ["from __future__ import annotations"]` — makes every annotation a lazy string, so modern spellings stay safe on 3.9

## Verification

- `ruff check` and `ruff format --check` clean; `ty check` clean; `pytest lens/tests` 148 passed
- Every lens module imports under KiCad's actual `python3.9.13` binary; the whole tree byte-compiles under it
- `cargo nextest run --workspace --profile ci --locked` — 1937 passed, 1 skipped
- **Proved the guard is real:** temporarily removing the future import from `lens/types.py` makes `cargo nextest run -p pcb-layout` fail with the Python import error. So the Rust suite genuinely drives these scripts through KiCad rather than passing vacuously — it is the only layer that would catch a 3.9 incompatibility.

---

> [!NOTE]
> **Low Risk**
> Lint and annotation/import hygiene only; sync logic is unchanged aside from documented noqa on existing silent exception paths.
> 
> **Overview**
> Adopts **ruff 0.16’s default lint set** for the KiCad layout Python tree (lens modules, tests, `update_layout_file.py`) after removing the prior pin to the older E/F-only selection. The diff is largely **mechanical cleanup**: PEP 585/604-style annotations, sorted imports, dropping redundant `.keys()` iteration, small refactors (e.g. collapsed `if`s, `ClassVar` on a mutable class default), and import reordering in `__init__.py`.
> 
> **KiCad runtime safety** is the substantive config change in `pyproject.toml`: **`target-version = "py39"`** so ruff does not assume 3.10+, plus **`extend-select = ["I002"]`** and **`required-imports = ["from __future__ import annotations"]`** so modern `list[str]` / `X | None` annotations stay valid under KiCad’s bundled **Python 3.9** (with comments explaining why dev-only pytest/ty on 3.12 won’t catch `|` at import time).
> 
> A handful of **broad `except` handlers** keep behavior but gain **`# noqa: BLE001`** (and FIXME notes where failures are silently swallowed in fragment discovery and connectivity rebuild).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a9ace0c2fc7b034014d6a525c19bbdba0383cd5a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/pcb-layout/src/scripts/lens/__init__.py
+++ b/crates/pcb-layout/src/scripts/lens/__init__.py
@@ -1,5 +1,7 @@
 """Lens-based layout synchronization module."""
 
-from .lens import run_lens_sync, SyncResult
+from __future__ import annotations
 
-__all__ = ["run_lens_sync", "SyncResult"]
+from .lens import SyncResult, run_lens_sync
+
+__all__ = ["SyncResult", "run_lens_sync"]
--- a/crates/pcb-layout/src/scripts/lens/changeset.py
+++ b/crates/pcb-layout/src/scripts/lens/changeset.py
@@ -10,22 +10,23 @@ Note: Renames (moved() paths) are handled in Rust preprocessing before
 the Python sync runs. This module no longer tracks renames.
 """
 
+from __future__ import annotations
+
 from dataclasses import dataclass, field
-from typing import Any, Dict, List, Literal, Optional, Set
+from typing import Any, Literal
 
 from .types import (
+    BoardComplement,
+    BoardView,
     EntityId,
     EntityPath,
-    Position,
-    BoardView,
-    BoardComplement,
-    FootprintView,
     FootprintComplement,
-    GroupView,
+    FootprintView,
     GroupComplement,
+    GroupView,
+    Position,
     default_footprint_complement,
 )
-
 
 # =============================================================================
 # Serialization Utilities
@@ -101,7 +102,7 @@ def parse_value(s: str) -> Any:
         return s
 
 
-def _split_list(s: str) -> List[str]:
+def _split_list(s: str) -> list[str]:
     """Split a comma-separated list, respecting quotes and brackets."""
     items = []
     current = ""
@@ -136,7 +137,7 @@ def _split_list(s: str) -> List[str]:
     return items
 
 
-def _tokenize(line: str) -> List[str]:
+def _tokenize(line: str) -> list[str]:
     """Tokenize a line into command and key=value pairs."""
     tokens = []
     current = ""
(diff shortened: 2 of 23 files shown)


akhilles:
### Bugbot: "Required-imports guard is inert" — valid, fixed in a9ace0c

Correct catch. `required-imports` is only enforced by `I002`, which is not in ruff's default selection (only `I001` is), so the setting was doing nothing and the comment above it claimed a safety net that did not exist. Now selected explicitly.

Verified both directions:

- **Stripping the future import from an existing file** now reports `I002` directly. This case was already caught indirectly — the file's existing modern annotations produce `F821`/`FA102` without it — so it was not the real exposure.
- **A brand-new file using only `Optional[...]`** now reports `I002`. *This was the genuine gap.* With no modern annotation syntax present, neither `F821` nor `FA102` fires, so such a file could have been added with no future import, and the first `X | None` annotation later added to it would have broken KiCad's Python 3.9.

### Build-perf comment: `demo/b/DM0001/Governor` 1.31× slower

Reading this as run-to-run variance, not a regression. This PR changes only Python sources under `crates/pcb-layout/src/scripts/` plus `pyproject.toml` lint config — no Rust code and no lockfile change. `pcb build` does not execute the layout-lens scripts at all (they run during `pcb layout`, through KiCad's interpreter), so there is no mechanism by which this diff could slow the build path. The other eight boards are flat within noise, and the reported figure is 90ms → 119ms with ±9ms stddev on a sub-100ms benchmark.

PR #1019, opened 2026-08-04
Add automatic copper balancing to the fab-panel panelizer

akhilles:
\`fab-panel create\` now balances copper automatically, the fabrication-panel pass that the board-array balancing doc reserved as out of scope.

- The balancing region is just the gutters: the usable stock area between the process margins, minus the placed assembly-panel outlines with the standard clearance, built and certified by the existing \`board_array_balancing_region\` construction. One safe region serves every copper layer, since the fab-panel step adds no per-layer geometry of its own.
- Each layer targets the aggregate copper density measured inside the placed panels, extending their already-balanced density into the gutters. The margins are excluded from the density domain entirely and stay bare.
- The joint spatial solve, stack weights, feature lowering, and per-layer summary/report are shared with board-array balancing via new crate-level \`copper_balance\` and \`generated\` modules (extracted from \`board_array\`, no behavior change); \`pcb-ir\` gains a \`collect_fab_panel_balancing_input\` collector.
- Generation is two-pass like board arrays: a provisional panel supplies composed copper and placed outlines, then the final panel is written with the generated copper as positive \`LayerFeature\` contours in the fab step.

Details in \`crates/pcb-ipc2581-tools/docs/fab-panel-copper-balancing.md\`.

Tests: gutters fill to the panel density with a forced-rotation placement, margins/clearances verified bare geometrically, full-coverage panel yields no fill, and sparse existing fixtures resolve to "no fill" so prior behavior is unchanged.

---

> [!NOTE]
> **Medium Risk**
> Touches fabrication-panel geometry, IPC negative sets, and Gerber polarity/export paths; changes are broad but covered by integration tests including Gerber–IPC area checks.
> 
> **Overview**
> **`fab-panel create` now runs automatic copper balancing** in the gutters between placed assembly panels (usable stock minus margins and panel outlines), targeting each layer’s density inside those panels. It uses the same two-pass flow as board arrays: provisional panel → solve → final XML with balance `LayerFeature`s and a per-layer summary on stderr.
> 
> **Board-array and fab-panel balancing share new `copper_balance` and `generated` modules** (logic moved out of `board_array` without intended behavior change). Perforated balance voids are emitted as a **positive plane plus negative `UserPrimitiveRef` instances** backed by `DictionaryUser` hex templates, instead of carving every full void into positive contours.
> 
> **`pcb-ir` adds `collect_fab_panel_balancing_input`**, keeps **clear polarity native** for artwork (no longer resolved in `normalize_for_artwork`), speeds set/cutout subtraction with bbox filtering, and tunes the spatial solver (more iterations, radius quantization step, shared safe-region caching, parallel uniform solves).
> 
> **Gerber export** optionally **resolves clear polarity geometrically** when sequential paint would disagree with layer-wide negatives; **repeated congruent clear regions** (≥16) collapse to shared outline-macro flashes. Fab-panel composition strips sources to manufacturing content up front and unions fabrication `sectionKey` across sources.
> 
> Docs: `fab-panel-copper-balancing.md` and changelog entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 48661eca7ea8278489a2d76e58715666c9f872b6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to Semantic Versioning (https://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- `fab-panel create` now automatically balances copper in the gutters between placed assembly panels.
+
 ### Fixed
 
 - Pad shapes with padstack offsets now land at the correct position on rotated pads across all IPC-2581 outputs.
--- a/crates/gerberx2/src/from_artwork.rs
+++ b/crates/gerberx2/src/from_artwork.rs
@@ -9,8 +9,8 @@ use std::collections::{BTreeSet, HashMap};
 
 use crate::{
     AttributeValue, Contour, ContourSegment, GerberError, GerberLayer, ObjectKind,
-    Point as GerberPoint, Result, WriterAperture, WriterApertureTemplate, WriterObject,
-    sanitize_attribute_field,
+    Point as GerberPoint, Result, WriterAperture, WriterApertureMacro, WriterApertureTemplate,
+    WriterMacroExpression, WriterMacroPrimitive, WriterObject, sanitize_attribute_field,
 };
 use pcb_ir::dialects::artwork::{Aperture, ApertureShape, Geometry as ArtworkGeometry, PaintStage};
 use pcb_ir::geom::path::{self as geom_path, ContourBuf, PathCmd};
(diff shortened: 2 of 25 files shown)


akhilles:
Follow-up commit: converges the spatial solve (radius fields were only ~half-amplified at 64 iterations), dedups shared safe regions, parallelizes the geometry passes, and restructures the fab-panel write pipeline (strip sources once, drop the redundant schema-validation pass). An 18×24 mixed fabrication panel generates in ~32 s instead of ~77 s; board arrays get the solver improvements too.

akhilles:
Addressed the three open review findings in 63fe6cf7: schema validation is restored on the final fabrication-panel document; clear-flash grouping uses an ordered map so macro names and aperture codes are deterministic; and clear features that can reach overlay-stage dark geometry (negative sets over pads/traces) fall back to layer-wide geometric polarity resolution, with a regression test pinning the composed image. Generated balance planes keep their voids native since they are certified clear of all other copper.

akhilles:
Both findings addressed in c29514e5: the clear-versus-overlay reachability check now hashes clear bounds into a uniform spatial grid (linear in feature count instead of clears × darks), and the translation-congruence signature includes cubic end points so curved clear shapes cannot collide onto one shared template.

akhilles:
Addressed in the latest commit: the clear-polarity reachability guard now also catches base-stage fills that follow a clear in source order (tracked incrementally against a spatial grid of preceding clears), with a regression test pinning the composed image of a negative set followed by an overlapping fill.

PR #1044, opened 2026-08-09
pcb-component-gen: detect pin name collisions before writing .zen file

Dev-X25874:
## What

When two distinct pin names in a KiCad symbol sanitize to the same
Starlark identifier (e.g. `~CS` and `!CS` both become `N_CS`), the
generated `.zen` file declares the same `io()` variable twice. This is
a Starlark error that previously only surfaced during `pcb build`
after the broken file had already been written to disk.

## Why

Silent failures here are especially bad in the registry/AI component
generation pipeline — a user runs `pcb search`, downloads a component,
and only finds out it's broken when they try to use it in a board.

## How

Added `check_pin_name_collisions()` in `pcb-component-gen` that inverts
the sanitized name map, detects any collisions, and returns a clear
error with the offending pin names and a fix hint — before any rendering
or file I/O happens.

Only applies to the symbol-backed path. The `explicit_pins` path is
caller-controlled so collision responsibility stays with the caller.

## Testing

- 3 new unit tests covering: `~`/`!` prefix collision, case collision,
  and a clean symbol that should pass through unchanged
- All 14 tests pass
- `cargo fmt` produced no changes

---

> [!NOTE]
> **Low Risk**
> Localized change to symbol-backed component `.zen` generation with deterministic suffixing and a focused unit test; explicit-pin generation is unchanged.
> 
> **Overview**
> **Component `.zen` generation** no longer maps distinct KiCad signal names that sanitize to the same Starlark identifier onto a single `io()` net (e.g. `~CS` and `!CS` both becoming `N_CS`).
> 
> `generated_signal_io_names` now runs sanitized names through **`uniquify_sanitized_names`**, which keeps the first signal on the base name and assigns **`_2`, `_3`, …** suffixes to later collisions in deterministic signal-name order. A unit test covers the `~CS` / `!CS` case.
> 
> **CHANGELOG** records this under Unreleased → Fixed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b23c54e321d295bc09ae7f9e586b872cc619dffb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ and this project adheres to Semantic Versioning (https://semver.org/spec/v2.0.0.
 - Pads with rounded or custom shapes now export as shared Gerber aperture flashes instead of repeated regions, shrinking mask and paste layers.
 - Rounded-rectangle pads export as compact parameterized macro apertures, and hexagon and octagon pads as polygon apertures.
 - Gerber output no longer repeats object attributes on every object.
+- Distinct KiCad signal names that sanitize to the same Starlark identifier are now uniquified (`N_CS`, `N_CS_2`, …) instead of silently merging onto one net.
 
 ## [0.4.25] - 2026-08-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to Semantic Versioning (https://semver.org/spec/v2.0.0.
 
 - BOM generation no longer panics when truncating UTF-8 component descriptions.
 - Nested modules retain group ownership for copied routing and graphics during layout synchronization.
+- Distinct KiCad signal names that sanitize to the same Starlark identifier are now uniquified instead of silently merging onto one net.
 
 ## [0.4.30] - 2026-08-14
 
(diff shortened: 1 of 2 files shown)


akhilles:
Thanks for looking into this! What prompted the change? Could you share the symbol and error you ran into? It looks like we already deduplicate the `io()` declarations, so I'm wondering if the actual issue is that both pins end up mapped to the same net.

Dev-X25874:
Thanks for the context! You're right that `pin_groups` is deduplicated via `BTreeSet`, so the `io()` declaration isn't actually duplicated.

The actual concern I was trying to address is in `pin_mappings` — which isn't deduplicated. If `~CS` and `!CS` both sanitize to `N_CS`, the generated `pins = {}` block ends up as:

```
pins = {
    "!CS": N_CS,
    "~CS": N_CS,
}

```

Both physical pins get mapped to the same net. If those are genuinely different pins with confusable names rather than two notations for the same signal, that's a silent short on the board.

Happy to close this if the consensus is that colliding names always mean the same signal and the existing behavior is intentional — just wanted to flag the case. Let me know how you'd like to handle it.

akhilles:
Instead of failing, could we uniquify them (`N_CS`, `N_CS_2`, etc.)?

Dev-X25874:
Done what you asked — uniquifying instead of failing now (`N_CS`, `N_CS_2`, etc). Tests pass, fmt clean.

akhilles:
the ci issue should be fixed on main, could you please rebase against main and move changelog entry to unreleased again? thanks!

akhilles:
Thanks!

PR #1088, opened 2026-08-25
ENG-1019 Cache PCB BOM matches per line

akhilles:
`pcb bom` needs reliable cached matches before the generic standard-library matching path is removed. This change adds one reusable write-through cache, stores each BOM line separately, refreshes stale or missing lines, and uses cached data when the API is unavailable.

---

> [!NOTE]
> **Medium Risk**
> Changes BOM sourcing cache semantics and partial API failure behavior, which affects availability output for `pcb bom` and offline workflows, though release checks now always hit the live API.
> 
> **Overview**
> **`pcb bom` availability matching** no longer caches one blob for the entire BOM request. Matches are stored **per BOM line** in a shared namespaced SQLite write-through cache (`bom-match-v2`), and online runs reuse fresh lines while calling the API only for stale or missing ones.
> 
> When a refresh fails, the CLI **merges whatever cached lines it still has** instead of requiring a full cached response. **`pcb bom --offline`** applies each cached line independently (any age); misses stay local-only without availability. Callers pass **`BomMatchOptions`** (mode plus configurable TTL, default ten minutes).
> 
> The old **`bom_cache`** module is removed in favor of generic **`WriteThroughCache`**. **`pcb publish` / release strict BOM offer checks** now use **`fetch_and_populate_availability_with_context`** (always live, no local match cache). Docs and the changelog describe the new per-line behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 18ab34183f9dda494ea1937280a6537c1d2d4062. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


akhilles:
bugbot run

PR #1090, opened 2026-08-25
Fix IPC DFM copper clearance ownership

akhilles:
IPC DFM treated same-net notches as copper-clearance defects. This change preserves conductor ownership through IPC artwork composition and checks clearance only between distinct conductors.


akhilles:
bugbot review

PR #1105, opened 2026-08-28
Support custom reference designator hints

### LINKED ISSUE #1057: Why is the list of known prefixes limited?
In crates/pcb-sch/src/lib.rs KNOWN_PREFIXES deliberately limits the prefixes that will be used to carry over stable names from .zen files to the PCB.  Why is this a whitelist instead of allowing any letter prefix?  In my PCB I have transformers marked T01 etc, and patching the source seems excessive just to allow those names to carry over.

akhilles:
Reference designator hints now accept any valid uppercase prefix that matches the component prefix. This removes the incomplete hard-coded whitelist while retaining existing ambiguity and collision safeguards. Fixes #1057.


mattclarkdotnet:
Awesome, thank you!

PR #1111, opened 2026-08-29
pcb-route-native: board->grid model for a native Rust autorouter

Dev-X25874:
﻿## What

New `pcb-route-native` crate: parses a `.kicad_pcb` file's footprints/pads/
Edge.Cuts outline and rasterizes them into a single-layer obstacle/keepout/
free grid. No pathfinding — this PR is scope-limited to the grid model only.

## Why

Groundwork for a native routing engine, to eventually remove `pcb route`'s
dependency on external FreeRouting (Java). Landing the grid model on its own
first so it's reviewable in isolation.

## Testing

- 6 unit/integration tests: footprint/pad parsing, rotated-footprint position
  resolution, Edge.Cuts outline extraction, keepout-radius boundary check,
  ASCII snapshot of a synthetic 2-footprint board
- `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean
- Verified footprint/pad counts against 8 real committed fixture boards
  (`crates/pcb-layout/tests/resources/*`, `crates/pcb-sch/test/kicad-bom`) -
  exact match against counts independently confirmed from the raw board text
- Explicit regression test/handling added for the zero-footprint case, found
  by running against `lib/std/test/layout/test_BoardConfig/layout.kicad_pcb`

## Known limitations

- None of the real boards checked in this repo have an actual `Edge.Cuts`
  graphic (only the layer *declaration*, never a drawn outline) - the
  outline-parsing path is only exercised by this PR's own synthetic fixture.
  Flagging rather than blocking; worth testing against a real board with a
  drawn outline before this is relied on.
- Footprint rotation sign convention is implemented and self-consistently
  tested, but not yet cross-checked against a real KiCad-exported rotated
  part.

---

> [!NOTE]
> **Low Risk**
> Self-contained new crate with no changes to existing routing or build paths; geometry uses approximations (pad circles, outline bbox) that may need validation before production routing.
> 
> **Overview**
> Introduces **`pcb-route-native`**, a new workspace library aimed at a future native Rust autorouter (replacing the FreeRouting dependency over time). This PR lands only the **geometry + rasterization** layer—**no pathfinding** and **no CLI wiring**.
> 
> **`parse_board`** reads `.kicad_pcb` via `pcb-sexpr`, extracting footprints/pads (with footprint rotation applied to pad positions), optional **Edge.Cuts** outline as a bounding box, and pad metadata (layers, nets). **`build_grid`** turns that into a single **`F.Cu`** `RoutingGrid` with **Free**, **Keepout** (clearance halo), and **Obstacle** cells, using conservative circular pad bounds and configurable pitch/clearance/margin. Boards without outline or pads get explicit fallback bounds (including a 1×1 grid for empty layouts).
> 
> Also adds an **`inspect_board`** example, synthetic fixture + **insta** ASCII grid snapshot, parsing/keepout tests, and an **[Unreleased]** changelog entry (prior unreleased fixes in the diff were removed from that section).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 13fc2f3f26b988dd99959d5cdaf37ca11b5d3227. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


akhilles:
sorry, but a new autorouter is way out of scope here and we can’t take on the maintenance burden. also, ai-assisted changes are welcome, but this is too large for us to review without a prior issue or discussion. please do keep working on it in a separate repo.

also any native auto-router that would be upstreamed into this project would prolly lean heavily on pcb-ir

PR #1116, opened 2026-08-30
Add self-contained JSON DFM reports with native geometry

akhilles:
DFM diagnostics need geometry and fabrication limits for visual inspection. Add one JSON report with native vector artwork, finding evidence, layout topology, and the exact PDK source so viewers need no companion files.


akhilles:
bugbot review

PR #1123, opened 2026-08-31
Make stdlib QR generic a white square

lucia3e8:
old QR code footprints look like this
image

which is out of spec with what JLC asks for 
image

we replace the corner-outline QR code footprint with a solid white square QR code footprint, because it's both easier for JLC to work with, and easier to laser etch if serializing by hand

new QR code footprints look like white squares that come in 5x5, 8x8 and 10x10mm variants
image
 
> [!NOTE]
> **Low Risk**
> Stdlib footprint and optional config only; no auth, routing, or BOM logic changes, with layout sync expected to preserve placement on FPID update.
> 
> **Overview**
> Updates the stdlib **`QR`** generic so JLCPCB’s “Specify Position” 2D barcode flow can find a matching **solid silkscreen square**, instead of the old hollow corner-bracket footprint that would not be replaced during fab.
> 
> **`QR.zen`** now documents the JLCPCB workflow, adds a **`size`** config (`5mm` / `8mm` / `10mm`, default `5mm`), sets component **`value`** / **`size`** properties, and picks the footprint via **`QR_<size>.kicad_mod`**. The single **`QR.kicad_mod`** is removed in favor of three footprints with filled **`F.SilkS`** rects at 5×5, 8×8, and 10×10 mm, plus fab outline, label, and +0.25 mm courtyard.
> 
> **`CHANGELOG.md`** records the change under Unreleased.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aea6836405bb6c28a96078dd5080891de1d94b0d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/std/generics/QR.zen
```diff
@@ -1,9 +1,23 @@
+Size = enum("5mm", "8mm", "10mm")
+
+size = config(Size, default=Size("5mm"), optional=True, help="2D barcode size")
```
akhilles: a bit silly but we support length as unit now so you can also do

```
size = config(Length, default="5mm", allowed=["5mm", "8mm", "10mm"], optional=True, help="2D barcode size")
```

i think

akhilles (approved):

akhilles (approved):

lib/std/generics/QR.zen
lucia3e8: that's awesome!

PR #1266, opened 2026-09-09
Add frame-only mouse-bite placement analysis

akhilles:
Add frame-only mouse-bite candidate placement and elastic analysis to select tab locations and counts. Use supplied courtyards, list ignored footprints without courtyards, and reject unusable courtyard geometry. Return conditional JSON analysis without generating tabs, holes, routes, or manufacturing geometry.


akhilles:
Superseded by #1269, which places tabs with a rigid-plate-on-tab-springs model instead of the plate FEM. The eligibility fixes from this branch were carried over there.

